PR #3, opened 2022-07-08
Get random cell data from BQ [VS-503]

rsasch:


src/random_bq_to_anndata.py
```diff
@@ -0,0 +1,46 @@
+from google.cloud import bigquery
+import argparse
+import random
+import anndata as ad
+
+
+def get_random_ids(project, dataset, num_cells):
+    client = bigquery.Client()
```
mcovarr, line 8: Can we create the client in one spot and pass it in to these two functions? Also George and I have sometimes found a need for the project to be explicitly specified a la

```
client = bigquery.Client(project=project)
```

src/random_bq_to_anndata.py
```diff
@@ -0,0 +1,46 @@
...
+    query = client.query(f"SELECT MAX(cas_cell_index) AS max_table_number FROM `{project}.{dataset}.cas_cell_info`")
+    max_cell_id = int([row.max_table_number for row in list(query.result())][0])
+    print(f"Getting {num_cells} random IDs between 0 and {max_cell_id}...")
+    cell_ids = list(range(0, max_cell_id))
```
mcovarr, line 12: `range` excludes its upper limit so
```suggestion
    cell_ids = list(range(0, max_cell_id + 1))
```

src/random_bq_to_anndata.py
```diff
@@ -0,0 +1,46 @@
...
+    random.shuffle(cell_ids)
+    del cell_ids[num_cells:]
```
mcovarr, line 14: TOL not sure if this or `cell_ids[0:num_cells]` is more efficient?

README.md
```diff
@@ -46,4 +44,12 @@ bq load -project_id broad-dsp-spec-ops -F tab --skip_leading_rows 1 kc_cas_test_
...
 ```
 
-### Extract Random Subset
\ No newline at end of file
+### Extract Random Subset
+
+This will randomly get a specified number of cells' data from your BigQuery dataset
+
+```
+python src/random_bq_to_anndata  --project <google-project> --dataset <dataset-name> --num_cells <num>
```
mcovarr, line 52:
```suggestion
python src/random_bq_to_anndata.py  --project <google-project> --dataset <dataset-name> --num_cells <num>
```

src/random_bq_to_anndata.py
```diff
@@ -0,0 +1,44 @@
...
+import argparse
+import random
+import anndata as ad
+
+
+def get_random_ids(project, dataset, client, num_cells):
+    query = client.query(f"SELECT MAX(cas_cell_index) AS max_table_number FROM `{project}.{dataset}.cas_cell_info`")
+    max_cell_id = int([row.max_table_number for row in list(query.result())][0])
+    print(f"Getting {num_cells} random IDs between 0 and {max_cell_id}...")
+    cell_ids = list(range(0, max_cell_id + 1))
```
mcovarr, line 11: This does assume the cell ids in the db blanket `[0, max_cell_id]` which is true for now, but if at some point that isn't the case we may get fewer results than requested. Might be worth adding a comment to this effect but I wouldn't bother coding for scenario unless that actually happens.

rsasch: added a comment

mcovarr (approved):

gbggrant (approved):

PR #7, opened 2022-07-14
VS-507. Read random data from BQ and convert to AnnData

gbggrant:
Working


src/random_bq_to_anndata.py
```diff
@@ -1,17 +1,20 @@
...
 import random
+import numpy as np
 import anndata as ad
-
+from scipy.sparse import csr_matrix
 
 # assumes that cas_cell_info.cas_feature_index values are a contiguous list of ints starting at 0
 def get_random_ids(project, dataset, client, num_cells):
-    query = client.query(f"SELECT MAX(cas_cell_index) AS max_table_number FROM `{project}.{dataset}.cas_cell_info`")
+    query = client.query(f"SELECT MIN(cas_cell_index) AS min_table_number, MAX(cas_cell_index) AS max_table_number FROM `{project}.{dataset}.cas_cell_info`")
```
mcovarr, line 10: I must have missed this on the last review; why are these columns selected as table numbers rather than cell indexes?

mcovarr: I'm beginning to wonder if we shouldn't just select the actual cell ids and randomize those... 🤔

src/random_bq_to_anndata.py
```diff
@@ -21,26 +24,105 @@ def get_cell_data(project, dataset, client, num_cells):
...
+    # AAC   0    n/a   4   n/a
+    # AAG   3    n/a   0   1
+    #
+    # we represent it as:
+    # index_ptr = [0, 3, 5, 8]
+    # indices   = [0, 1, 2, 0, 2, 0, 2, 3]
+    # data      = [1, 7, 2, 0, 4, 3, 0, 1]
+
+    last_original_cell_id = None
     for row in list(cell_data):
```
mcovarr, line 68: if the return from `get_cell_data` is iterable it might be preferable to iterate that rather than force evaluation to a list

src/random_bq_to_anndata.py
```diff
@@ -21,26 +24,105 @@ def get_cell_data(project, dataset, client, num_cells):
...
+            feature_ids.append(original_feature_id)
+            feature_names.append(row["feature_name"])
+            column_count += 1
+        col_num = feature_to_column[original_feature_id]
+
+        indices.append(col_num)
+        data.append(count)
+
+#TODO - catch the edge case if the last row is different from last-1
+# And add a test - especially for last-1
```
mcovarr, line 98: yes please, this is tricky code 😄

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+    return ordered_cell_indices, cell_index_to_cell
+
+def get_matrix_data(project, dataset, client, random_cell_ids):
+    in_clause = f" matrix.cas_cell_index IN ({','.join(map(str, random_cell_ids))})"
 
     # at some point, we will probably want create temp table of cell_ids and then JOIN on it
     # instead of an IN clause
-    sql = f"SELECT original_cell_id, cell_type, original_feature_id, feature_name, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_feature_info` AS feature, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND matrix.cas_feature_index = feature.cas_feature_index AND" + in_clause
-    print(f"Getting {num_cells} random cells' data from {project}.{dataset}...")
+    # NOTE - really don't need to join to cas_cell_index and cas_feature_index here anymore.
```
mcovarr, line 70: any reason not to remove these joins?

src/random_bq_to_anndata.py
```diff
@@ -21,26 +24,105 @@ def get_cell_data(project, dataset, client, num_cells):
 
     # at some point, we will probably want create temp table of cell_ids and then JOIN on it
     # instead of an IN clause
-    sql = f"SELECT original_cell_id, cell_type, original_feature_id, feature_name, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_feature_info` AS feature, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND matrix.cas_feature_index = feature.cas_feature_index AND" + in_clause
+    sql = f"SELECT matrix.cas_cell_index, original_cell_id, cell_type, matrix.cas_feature_index, original_feature_id, feature_name, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_feature_info` AS feature, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND matrix.cas_feature_index = feature.cas_feature_index AND" + in_clause + " ORDER BY matrix.cas_cell_index, matrix.cas_feature_index"
     print(f"Getting {num_cells} random cells' data from {project}.{dataset}...")
```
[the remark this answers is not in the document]
gbggrant: Not sure what gaps you mean - the cells?

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+    for row in query:
+        index = row["cas_feature_index"]
+        ordered_feature_indices.append(index)
+        feature_index_to_feature[index] = Feature(row["cas_feature_index"], row["original_feature_id"], row["feature_name"])
+    return ordered_feature_indices, feature_index_to_feature
 
-def get_cell_data(project, dataset, client, num_cells):
-    random_ids = get_random_ids(project, dataset, client, num_cells)
-    in_clause = f" matrix.cas_cell_index IN ({','.join(map(str, random_ids))})"
+# Retrieve metadata for the selected (random) cells from the database
```
mcovarr, line 52: Not sure how we're going to end up defining "metadata" for this project

```suggestion
# Retrieve data for the selected (random) cells from the database
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
-    cell_data = get_cell_data(project, dataset, client, num_cells)
-    adata = ad.AnnData()
-    for row in list(cell_data):
-        # insert VS-507 here
-        pass
+
+    print(f"Getting {num_cells} random cells' data from {project}.{dataset}...")
+    random_cell_ids = get_random_cell_ids(project, dataset, client, num_cells)
+
+    # Read the cell information and store metadata for later
```
mcovarr, line 82:
```suggestion
    # Read the cell information and store for later
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+    ordered_cell_indices, cell_index_to_cell = get_cells(project, dataset, client, random_cell_ids)
+
+    original_cell_ids = []
+    cell_types = []
+    for cell_index in ordered_cell_indices:
+        original_cell_ids.append(cell_index_to_cell[cell_index].original_cell_id)
+        cell_types.append(cell_index_to_cell[cell_index].cell_type)
+
+    # Read the feature information and store metadata for later.
```
mcovarr, line 91:
```suggestion
    # Read the feature information and store for later.
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+    ordered_feature_indices, feature_index_to_feature = get_features(project, dataset, client)
+
+    feature_ids = []
+    feature_names = []
+    last_feature_index = None
+    for feature_index in ordered_feature_indices:
+        if (last_feature_index is None):
+            last_feature_index = feature_index - 1
+        # # We expect all entries in cas_feature_info to be stored in continuous fashion. (1, 2, 3, ...)
```
mcovarr, line 100:
```suggestion
        # We expect all entries in cas_feature_info to be stored in continuous fashion. (1, 2, 3, ...)
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+        if (feature_index != last_feature_index + 1):
```
mcovarr, line 101:
```suggestion
        if feature_index != last_feature_index + 1:
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+    adata = ad.AnnData(counts)
+    adata.obs_names = original_cell_ids
+    adata.obs["cell_type"] = cell_types
+    adata.var_names = feature_ids
+    adata.var["feature_name"] = feature_names
+
+    # See https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.raw.html?highlight=raw#anndata.AnnData.raw
+    #  for why we set 'raw' thusly: "The raw attribute is initialized with the current content of an object by setting:"
+    adata.raw = adata
+    adata.write(f'{output_file_prefix}.h5ad', compression="gzip")
```
mcovarr, line 124: PyCharm is not happy with this, it wants the first arg to be a `PathLike` rather than a string

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+        cell_types.append(cell_index_to_cell[cell_index].cell_type)
+
+    # Read the feature information and store metadata for later.
+    ordered_feature_indices, feature_index_to_feature = get_features(project, dataset, client)
+
+    feature_ids = []
+    feature_names = []
+    last_feature_index = None
+    for feature_index in ordered_feature_indices:
+        if (last_feature_index is None):
```
mcovarr, line 98:
```suggestion
        if last_feature_index is None:
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+
+    # Note that this method requires that the result set returned by get_cell_data be sorted by cas_cell_index (or, could also be sorted by original_cell_id)
+    cell_data = get_matrix_data(project, dataset, client, random_cell_ids)
+
+    (index_ptr, indices, data) = generate_sparse_matrix(cell_data, ordered_feature_indices[0])
+
+    # Create the matrix from the sparse data representation generated above.
+    counts = csr_matrix((data, indices, index_ptr), dtype=np.float32)
+    adata = ad.AnnData(counts)
+    adata.obs_names = original_cell_ids
```
mcovarr, line 116: looks like our existing code refers to a similar concept as `adata.obs.index` and `adata.var.index`, is there a difference?

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
-    adata = ad.AnnData()
-    for row in list(cell_data):
-        # insert VS-507 here
-        pass
+
+    print(f"Getting {num_cells} random cells' data from {project}.{dataset}...")
+    random_cell_ids = get_random_cell_ids(project, dataset, client, num_cells)
+
+    # Read the cell information and store metadata for later
+    ordered_cell_indices, cell_index_to_cell = get_cells(project, dataset, client, random_cell_ids)
```
mcovarr, line 83: could this just return ordered `Cell`s instead?

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+
+    original_cell_ids = []
+    cell_types = []
+    for cell_index in ordered_cell_indices:
+        original_cell_ids.append(cell_index_to_cell[cell_index].original_cell_id)
+        cell_types.append(cell_index_to_cell[cell_index].cell_type)
+
+    # Read the feature information and store metadata for later.
+    ordered_feature_indices, feature_index_to_feature = get_features(project, dataset, client)
```
mcovarr, line 92: similar question here as above with returning ordered `Feature`s

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,178 @@
...
+
+    feature_ids = []
+    feature_names = []
+    last_feature_index = None
+    for feature_index in ordered_feature_indices:
+        if (last_feature_index is None):
+            last_feature_index = feature_index - 1
```
mcovarr, line 99: this sets `last_feature_index` to -1 in this edge case?

src/test_random_bq_to_anndata.py
```diff
@@ -0,0 +1,110 @@
+import unittest
+import numpy
+
+from random_bq_to_anndata import generate_sparse_matrix
+
+class TestRandomBqToAnndata(unittest.TestCase):
```
mcovarr, line 6:
```suggestion
class TestRandomBqToAnnData(unittest.TestCase):
```

src/test_random_bq_to_anndata.py
```diff
@@ -0,0 +1,110 @@
...
+    def test_one_cell_sparse_matrix(self):
```
mcovarr, line 7: technically it's 100% dense... 😛

src/random_bq_to_anndata.py
gbggrant: Yup, good idea.

src/random_bq_to_anndata.py
gbggrant: Yup, good idea.

src/random_bq_to_anndata.py
gbggrant: Which is fine, then I still can get that it's a continuous list (-1, 0, 1, 2). But I've rewritten it and put the logic into the get_features method.

src/test_random_bq_to_anndata.py
gbggrant: This is true.

src/random_bq_to_anndata.py
gbggrant: I've removed the join on features. I'm trying to convince myself that it's safe to remove that to cell. But we *could* have a case where one of the cas_cell_ids is NOT in the matrix.

Probably going to rewrite it to have the random ids be pulled from the matrix itself.

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,167 @@
...
 import argparse
 import random
+import numpy as np
 import anndata as ad
+from scipy.sparse import csr_matrix
 
+class Cell:
+    cas_cell_index = 0
+    original_cell_id = ""
+    cell_type = ""
```
mcovarr, line 11: not required

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,167 @@
...
+        if (len(features) > 0) and (feature.cas_feature_index != features[-1].cas_feature_index + 1):
+            # It's not the first element in the list, and it's cas_feature_index is not one more than the previous element's
+            raise Exception("ERROR: Non-continuous values for `cas_feature_index` in table `cas_feature_info`")
+        features.append(feature)
+    return features
+
+# Return a list of cell objects (for the random_cell_ids), ordered by cas_cell_index
+def get_cells(project, dataset, client, cell_ids):
+    in_clause = f" cas_cell_index IN ({','.join(map(str, cell_ids))})"
+    sql = f"SELECT cas_cell_index, original_cell_id, cell_type FROM `{project}.{dataset}.cas_cell_info` WHERE " + in_clause + " ORDER BY cas_cell_index"
```
mcovarr, line 56:
```suggestion
    sql = f"SELECT cas_cell_index, original_cell_id, cell_type FROM `{project}.{dataset}.cas_cell_info` WHERE {in_clause} ORDER BY cas_cell_index"
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,167 @@
...
+def get_matrix_data(project, dataset, client, random_cell_ids):
+    in_clause = f" matrix.cas_cell_index IN ({','.join(map(str, random_cell_ids))})"
 
     # at some point, we will probably want create temp table of cell_ids and then JOIN on it
     # instead of an IN clause
-    sql = f"SELECT original_cell_id, cell_type, original_feature_id, feature_name, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_feature_info` AS feature, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND matrix.cas_feature_index = feature.cas_feature_index AND" + in_clause
-    print(f"Getting {num_cells} random cells' data from {project}.{dataset}...")
+    # NOTE - really don't need to join to cas_cell_index and cas_feature_index here anymore.
+    sql = f"SELECT matrix.cas_cell_index, original_cell_id, matrix.cas_feature_index, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND" + in_clause + " ORDER BY matrix.cas_cell_index, matrix.cas_feature_index"
```
mcovarr, line 69:
```suggestion
    sql = f"SELECT matrix.cas_cell_index, original_cell_id, matrix.cas_feature_index, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND {in_clause} ORDER BY matrix.cas_cell_index, matrix.cas_feature_index"
```

mcovarr (commented):
some commented out debug code to be removed, otherwise looking close!

src/random_bq_to_anndata.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, cas_feature_index, original_feature_id, feature_name):
         self.original_feature_id = original_feature_id
         self.feature_name = feature_name
 
-# assumes that cas_cell_info.cas_cell_index values are a contiguous list of ints
+# assumes that cas_cell_info.cas_cell_index values are a continuous list of ints
```
kcibul, line 28: I can't comment below… but in this line `list(range(min_cas_cell_index, max_cas_cell_index + 1))` just consider much memory this will consume for a bigger dataset?  Might be fine

src/random_bq_to_anndata.py
kcibul: Pulling from the matrix is going to be extremely expensive… It's valid (though strange) to have a cell_id that has no row in the matrix.  It means that all the values were zero (since this is a sparse matrix).  We should never have a value in the matrix thought that does't have a corresponding entry in the cell table.

src/random_bq_to_anndata.py
```diff
@@ -112,7 +112,7 @@ def random_bq_to_anndata(project, dataset, num_cells, output_file_prefix):
     adata.raw = adata
     adata.write(f'{output_file_prefix}.h5ad', compression="gzip")
 
-def generate_sparse_matrix(cell_data, minimum_feature_index):
+def generate_sparse_matrix(cell_data, minimum_feature_index, maximum_feature_index):
```
kcibul, line 115: I think you might be able to do this in COO format more easily… it's exactly what we're storing in BQ as opposed to a CSR format (which is complex as you're seeing!)

src/random_bq_to_anndata.py
```diff
@@ -66,7 +66,7 @@ def get_matrix_data(project, dataset, client, random_cell_ids):
     # at some point, we will probably want create temp table of cell_ids and then JOIN on it
     # instead of an IN clause
     # NOTE - really don't need to join to cas_cell_index and cas_feature_index here anymore.
-    sql = f"SELECT matrix.cas_cell_index, original_cell_id, matrix.cas_feature_index, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND" + in_clause + " ORDER BY matrix.cas_cell_index, matrix.cas_feature_index"
+    sql = f"SELECT matrix.cas_cell_index, matrix.cas_feature_index, raw_counts AS count FROM `{project}.{dataset}.cas_cell_info` AS cell, `{project}.{dataset}.cas_raw_count_matrix` AS matrix WHERE matrix.cas_cell_index = cell.cas_cell_index AND" + in_clause + " ORDER BY matrix.cas_cell_index, matrix.cas_feature_index"
```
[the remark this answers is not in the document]
gbggrant: I'm pulling all of the cell's data directly from cas_cell_index and storing the original_cell_id's as `obs_names` later.

src/random_bq_to_anndata.py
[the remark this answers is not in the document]
gbggrant: Nice (and thank you!). I have used suggestion 2.

gbggrant: Since I've re-implemented using the in-database randomization/selection than this point is no longer relevant.

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,144 @@
...
+    cells = get_cells(project, dataset, client, random_cell_ids)
+    original_cell_ids = []
+    cell_types = []
+    cas_cell_index_to_row_num = {}
+    row_num = 0
+    for cell in cells:
+        original_cell_ids.append(cell.original_cell_id)
+        cell_types.append(cell.cell_type)
+        cas_cell_index_to_row_num[cell.cas_cell_index] = row_num
+        row_num += 1
```
mcovarr, line 78: it looks like [enumerate](https://docs.python.org/3/library/functions.html#enumerate) would provide the index variable without needing to create one explicitly
```suggestion
    for row_num, cell in enumerate(cells):
        original_cell_ids.append(cell.original_cell_id)
        cell_types.append(cell.cell_type)
        cas_cell_index_to_row_num[cell.cas_cell_index] = row_num
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,144 @@
...
+
+    feature_ids = []
+    feature_names = []
+    cas_feature_index_to_col_num = {}
+    col_num = 0
+    for feature in features:
+        feature_ids.append(feature.original_feature_id)
+        feature_names.append(feature.feature_name)
+        cas_feature_index_to_col_num[feature.cas_feature_index] = col_num
+        col_num += 1
```
mcovarr, line 91: here too
```suggestion
    for col_num, feature in enumerate(features):
        feature_ids.append(feature.original_feature_id)
        feature_names.append(feature.feature_name)
        cas_feature_index_to_col_num[feature.cas_feature_index] = col_num
```

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,144 @@
...
+    adata.obs["cell_type"] = cell_types
+    adata.var_names = feature_ids
+    adata.var["feature_name"] = feature_names
+
+    # See https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.raw.html?highlight=raw#anndata.AnnData.raw
+    #  for why we set 'raw' thusly: "The raw attribute is initialized with the current content of an object by setting:"
+    adata.raw = adata
+    adata.write(f'{output_file_prefix}.h5ad', compression="gzip")
+
+def generate_coo_sparse_matrix(cell_data, cas_cell_index_to_row_num, cas_feature_index_to_col_num):
```
mcovarr, line 112: naming nitpick, but this doesn't actually include generating the coo sparse matrix

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,144 @@
...
+    cas_feature_index_to_col_num = {}
+    col_num = 0
+    for feature in features:
+        feature_ids.append(feature.original_feature_id)
+        feature_names.append(feature.feature_name)
+        cas_feature_index_to_col_num[feature.cas_feature_index] = col_num
+        col_num += 1
+
+    # Note that this method requires that the result set returned by get_cell_data be sorted by cas_cell_index (or, could also be sorted by original_cell_id)
+    cell_data = get_matrix_data(project, dataset, client, random_cell_ids)
```
mcovarr, line 94: the variable name is kind of confusing, maybe  `matrix_data` or `raw_data`?

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,144 @@
...
+            raise Exception(f"ERROR: Unable to find entry for cas_cell_index: {cas_cell_index} in lookup table")
+
+        try:
+            col_num = cas_feature_index_to_col_num[cas_feature_index]
+        except KeyError:
+            raise Exception(f"ERROR: Unable to find entry for cas_feature_index: {cas_feature_index} in lookup table")
+
+        rows.append(row_num)
+        columns.append(col_num)
+        data.append(row["count"])
```
mcovarr, line 132: nice, this is a lot easier to follow than before 😅

src/random_bq_to_anndata.py
gbggrant: Done

src/random_bq_to_anndata.py
gbggrant: Yup, I've changed the name.

src/random_bq_to_anndata.py
gbggrant: Yup, I've changed the name.

src/random_bq_to_anndata.py
gbggrant: 👍

src/random_bq_to_anndata.py
gbggrant: I've removed the joins. The code handles this case (where a cell that is present in cas_cell_index has no data in the matrix).

src/random_bq_to_anndata.py
gbggrant: Okay, I've made the first arg a `Path`- don't have PyCharm running on my machine, so let me know if it's still not happy.

src/random_bq_to_anndata.py
gbggrant: Really not sure. I've changed the code to use `adata.obs.index` and `adata.var.index` for consistency. They *seem* to populate the `adata.obs_names` and `adata.var_names` in the anndata structure when I browse it.

src/random_bq_to_anndata.py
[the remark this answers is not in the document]
gbggrant: I've removed the (unneeded) constraint about there not being a gap. The method now returns an ordered list of objects - tuples might be more memory efficient, but having the object makes it easier to populate more feature-specific metadata (that is pull things other than feature_name when/if it becomes necessary)

src/random_bq_to_anndata.py
gbggrant: Yes, thanks - I've done that. Much simpler. A shame I didn't do that *before* getting it all working using CSR.

src/random_bq_to_anndata.py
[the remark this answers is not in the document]
gbggrant: The code changes associated with going to coo_matrix should obviate this.

src/random_bq_to_anndata.py
```diff
@@ -1,46 +1,136 @@
...
-    return cell_ids
+class Feature:
+    def __init__(self, cas_feature_index, original_feature_id, feature_name):
+        self.cas_feature_index = cas_feature_index
+        self.original_feature_id = original_feature_id
+        self.feature_name = feature_name
 
+# Returns an array of <num_cells> cas_cell_index values pulled at random from the cas_cell_info table
+def get_random_cell_ids(project, dataset, client, num_cells):
+    query = client.query(f"SELECT cas_cell_index, rand() AS rand_val FROM `{project}.{dataset}.cas_cell_info` ORDER BY rand_val LIMIT {num_cells}")
```
mcovarr, line 22: nice!

mcovarr (approved):
LGTM! 👍

src/random_bq_to_anndata.py
gbggrant: I can't take credit for that!  ( @kcibul )

PR #10, opened 2022-07-21
Ingest and extract obs, var, and uns metadata [VS-533] [VS-534] [VS-536] [VS-537]

mcovarr:
Store and extract obs and var metadata as JSON strings in cells and features respectively.

### DIFF
--- a/src/anndata_to_avro.py
+++ b/src/anndata_to_avro.py
@@ -171,11 +171,17 @@ def dump_ingest_info(adata, filename, ingest_id):
     parsed_schema = parse_schema(schema)
 
     def sanitize_uns():
+        """
+        AnnData `uns` sometimes contains numpy arrays which the json library does not know how to serialize.
+        Be on the lookout for these and if we find any convert them to regular Python arrays.
+        :return: `uns` sanitized of numpy `ndarray`s.
+        """
         ret = {}
         for k, v in adata.uns.items():
             if isinstance(v, (np.ndarray,)):
                 v = v.tolist()
             ret[k] = v
+        return ret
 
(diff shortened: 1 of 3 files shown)


src/anndata_to_avro.py
```diff
@@ -109,10 +112,9 @@ def cell_generator():
 
 
 def dump_feature_info(adata, filename, cas_feature_index_start, ingest_id):
-    # Dump out feature info -- feature_id is index (12 columns).
```
mcovarr, line 112: not universally true

src/load_dataset.py
```diff
@@ -79,6 +85,9 @@ def create_bigquery_objects(client, project, dataset):
                      bigquery.SchemaField("cas_feature_index", "INTEGER", mode="REQUIRED"),
                      bigquery.SchemaField("original_feature_id", "STRING", mode="REQUIRED"),
                      bigquery.SchemaField("feature_name", "STRING", mode="REQUIRED"),
+                     bigquery.SchemaField("var_metadata", "STRING", mode="REQUIRED"),
+                     # TODO get direct JSON metadata loading working
```
gbggrant, line 89: Is there a ticket for this?

gbggrant (approved):

src/load_dataset.py
mcovarr: No but there should be. Kristian and I talked about it briefly in Slack on Friday afternoon and I had hoped to roll it into this PR but that didn't work out.

mcovarr: VS-546 created.

PR #11, opened 2022-07-26
Cap `uns` metadata entry size [VS-547]

mcovarr:


src/anndata_to_avro.py
```diff
@@ -209,8 +209,17 @@ def default(self, obj):
             return json.JSONEncoder.default(self, obj)
 
     def ingest_generator():
+        # Some `uns` metadata has extremely large values that frustrate extract. Cap the allowed size and substitute
+        # a `None` value in these cases.
+        metadata_limit = 2 ** 20
```
gbggrant, line 214: Any particular reason for this cap?

src/anndata_to_avro.py
```diff
@@ -209,8 +209,17 @@ def default(self, obj):
...
+        uns = {}
+        for k, uncapped_v in adata.uns.data.items():
+            j = json.dumps(uncapped_v, cls=NumpyEncoder)
+            v = None if len(j) > metadata_limit else uncapped_v
```
gbggrant, line 218: Would it be worth logging the type of metadata (is that available) that you are capping so that we understand if there are certain types that we can just exclude (rather than exclude purely by size)

gbggrant (approved):
LGTM

src/anndata_to_avro.py
mcovarr: tl;dr big but not too big. I added some prose rationalizing the choice.

PR #12, opened 2022-07-28
Ingest / extract multiple AnnDatas to / from a single dataset [VS-541]

mcovarr:


README.md
```diff
@@ -19,13 +19,21 @@ And a larger data set
 
 ### Convert AnnData to Avro Load Format
 
-First convert from AnnData to the Avro format that will be used to load BigQuery:
+First convert from AnnData to the Avro format that will be used to load BigQuery. The `anndata_to_avro.py` script needs
+to know the appropriate values to use for cell and feature indexes. These can either be specified explicitly as command
+line arguments, or the script can be pointed to an existing dataset / project to figure those out itself.
+
+With explicit start values:
```
kcibul, line 26: I would flip the order of these, and make it clear the explicit start values is a "Note: if you need to explicitly…".  Your new way is much nicer!

src/anndata_to_avro.py
```diff
@@ -299,16 +342,28 @@ def process(input_file, cas_cell_index_start, cas_feature_index_start, avro_pref
     parser.add_argument('--avro_prefix', type=str,
                         help='Prefix to use for output Avro files, e.g. <avro_prefix>_cell_info.avro etc',
                         required=False)
-    parser.add_argument('--cas_cell_index_start', type=int, help='starting number for cell index', required=False,
-                        default=0)
-    parser.add_argument('--cas_feature_index_start', type=int, help='starting number for feature index', required=False,
-                        default=0)
+    parser.add_argument('--cas_cell_index_start', type=int, help='starting number for cell index', required=False)
```
kcibul, line 345: Love this change!

src/random_bq_to_anndata.py
```diff
@@ -112,55 +118,62 @@ def random_bq_to_anndata(project, dataset, num_cells, output_file_prefix):
...
-    # for why we set 'raw' thusly: "The raw attribute is initialized with the current content of an object by setting:"
-    adata.raw = adata
-    adata.write(Path(f'{output_file_prefix}.h5ad'), compression="gzip")
-
-
-def convert_matrix_data_to_coo_matrix_input_format(cell_data, cas_cell_index_to_row_num, cas_feature_index_to_col_num):
+    random_cells = get_random_cells(project, dataset, client, num_cells)
+
+    cells_by_ingest_id = defaultdict(list)
+    for cell in random_cells:
```
kcibul, line 124: neat, very clean

kcibul (approved):
Nice implementation.  The only thing nagging at me is the holding of the entire count matrix in memory before writing it out.  Not new in this PR, and likely fine at the scale of 10k cells so good for now!

PR #13, opened 2022-08-01
Add basic unit test CI, linting [VS-550]

mcovarr:
Happy to do a walk through of this. The changes are not as large as they would appear; there's a lot of file moving-around and automated reformatting that greatly exaggerates the diffs.

The basic structure is largely "inspired" by (i.e. copied from) the fine work done in [Cromshell 2.0](https://github.com/broadinstitute/cromshell/tree/cromshell_2.0). The [tox](https://tox.wiki/en/latest/) system is used to drive format checking, linting, and test execution by `pylint`, `black`, `isort` and `flake8`.  As part of incorporating `tox`, casp was module-ized which required moving existing files around and creating a bunch of empty `__init__.py` files.

Some of the format checking / linting (e.g. import ordering) will require fixes; issues like this will produce non-zero exit codes from the identifying tools that will fail the associated GH Action. Fortunately issues like `isort` import ordering are easy to fix manually and then shouldn't need to be touched again. More substantive formatting issues like the things `black` complains about can be fixed automatically with

```shell
tox -e format
```

`pylint` is being invoked with the `--exit-zero` option which makes its recommendations optional, though I did incorporate a lot of its suggestions to improve the score of this code base. 🙂


rsasch (commented):
I think you might need to update the command paths in the README.md to reflect the addition of the `casp` directory

rsasch (approved):
Looks good!  One nit: please change the "gcs_prefix" parameter in `load_dataset.py` to be something clearer like "gcs_path_prefix".

src/casp/anndata_to_avro.py
```diff
@@ -324,7 +362,8 @@ def process(input_file, cas_cell_index_start, cas_feature_index_start, avro_pref
     print("Processing core data...")
     # recode the indexes to be the indexes of obs/var or should obs/var include these indices?
     dump_core_matrix(adata.raw.X, row_index_to_cas_cell_index, col_index_to_cas_feature_index, raw_counts_filename)
-        
+
+
```
gbggrant, line 366: Is this commented block still needed?

gbggrant (approved):

src/casp/anndata_to_avro.py
mcovarr: If I understand this correctly then I think so. I believe this is asking for an ingest-side test to make sure the representations we're creating are consistent with the contents of the AnnData file (there are currently no tests for ingest to parallel the ones you created for extract).

kcibul (approved):
The good thing about being late is that everyone else commented first.  Great work, really nice to have this foundation in here.

PR #14, opened 2022-08-10
Use JSON columns for metadata [VS-546]

mcovarr:
Per discussion with Google, the previous situation of being unable to load JSON columns from Avro files is a bug which Google was aware of and for which a fix is currently being rolled out. The fix is expected to become generally available in one to two weeks, but in the meantime projects must be put on an "allow list" in order for JSON column loading from Avro files to work. Our `gvs-internal` project has been allowlisted so this code works fine there, but `broad-dsp-spec-ops` has not been allowlisted so this code does not work there.

### DIFF
--- a/src/casp/anndata_to_avro.py
+++ b/src/casp/anndata_to_avro.py
@@ -108,7 +108,7 @@ def dump_cell_info(adata, filename, cas_cell_index_start, ingest_id):
             {"name": "cas_cell_index", "type": "int"},
             {"name": "original_cell_id", "type": "string"},
             {"name": "cell_type", "type": "string"},
-            {'name': 'obs_metadata', 'type': {'type': 'string', 'sqlType': 'JSON'}},
+            {"name": "obs_metadata", "type": {"type": "string", "sqlType": "JSON"}},
             {"name": "cas_ingest_id", "type": "string"},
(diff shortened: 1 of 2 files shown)


src/casp/anndata_to_avro.py
```diff
@@ -28,7 +28,7 @@ def current_milli_time():
     return round(time.time() * 1000)
 
 
-def write_avro(generator, parsed_schema, filename, progress_batch_size=1000):
+def write_avro(generator, parsed_schema, filename, progress_batch_size=10000):
```
rsasch, line 31: just curious — why was this upped?

rsasch (approved):
One question, but looks good 👍🏻

src/casp/anndata_to_avro.py
mcovarr: I apparently had an issue building a native code piece of a numerical library for one of my Python versions, which caused the library to fall back to a super slow pure Python implementation. At some point I reinstalled this Python version but this time the library successfully. These operations got ~10x faster and I was able to raise the size of batches between progress updates.

mcovarr:
JSON column loading from AVRO is GA now, merging.

PR #20, opened 2022-11-15
Implement Embeddings + restructure services

fedorgrab:


kcibul (approved):
Looks like mostly package refactoring and extracting common methods, right?  If there's something deeper let me know so I can take a better look!

fedorgrab:
> Looks like mostly package refactoring and extracting common methods, right? If there's something deeper let me know so I can take a better look!

Yeah, mostly what you said. I also added automated embeddings.

PR #27, opened 2022-11-21
Chore: Add 8 workers to inference and api servers

fedorgrab:


src/casp/api/server.py
```diff
@@ -166,4 +166,4 @@ async def annotate(myfile: UploadFile):
 
 
 if __name__ == "__main__":
-    uvicorn.run("server:app", host=settings.server_host, port=settings.server_port)
```
kcibul, line 169: Unless you've done some specific tuning, maybe make this default dynamic following [this recommendation](https://stackoverflow.com/questions/72374634/how-many-uvicorn-workers-do-i-have-to-have-in-production) of 2 * vCPUs + 1

```
import multiprocessing

workers=multiprocessing.cpu_count()* 2 + 1
```

kcibul (approved):
Looks good - one minor comment.  I also think we should change the `requirements.txt` to install `uvicorn[standard]` so we get `uvloop` and other performance boosts out of the box.

PR #33, opened 2022-11-28
Chore: Include isort to tox format

fedorgrab:


kcibul (approved):
but also add the isort line length

PR #43, opened 2022-12-07
Chore: #40 Add docker container update to CI/CD

fedorgrab:


.github/workflows/release_workflow.yml
```diff
@@ -0,0 +1,104 @@
+name: CAS Release Workflow
+
+# Controls when the action will run.
+# Triggers the workflow on push or pull request events
+on: [push, workflow_dispatch]
```
kcibul, line 5: This seems like it runs on any push or whenever another workflow runs?  But that doesn't match with the comment.

kcibul, line 5: Similarly if this runs on every push to any branch… won't all these docker images clobber eachother?  For example if user A and user B both have their own feature branches and push to those branches.. they will both push docker images to `:latest`

Typically I've seen docker building only happy on a push to `main` or some other protected branch

kcibul: reading more below it looks like it's not pushing for any commit but only a tagged commit, but that could still conflict right? Might be better to have it on merge to `main` or on the creation of a release (see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release)

.github/workflows/release_workflow.yml
```diff
@@ -0,0 +1,104 @@
...
+  artifact-update:
+    name: Google Artifact Registry container update
+    runs-on: ubuntu-latest
+    needs: [tox-python-build]
+    env:
+      DOCKER_REGISTRY_NAME: us-central1-docker.pkg.dev
+      PYTORCH_IMAGE_NAME: us-central1-docker.pkg.dev/${{ secrets.GCP_PROJECT_ID }}/cas-services-cicd/cas-pytorch
+      PYTORCH_CUDA_IMAGE_NAME: us-central1-docker.pkg.dev/${{ secrets.GCP_PROJECT_ID }}/cas-services-cicd/cas-pytorch-cuda
+    # Notice it's running only when a tagged commit is pushed
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
kcibul, line 53: Why not just make this a separate workflow triggered by the event we want?  And then leaving unit_test.yml as is (was?)

.github/workflows/release_workflow.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    steps:
+      - id: checkout
+        name: Checkout
+        uses: actions/checkout@v2
+
+      - id: get-tag
+        # Retrieving git release tag
+        name: Get tag
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
```
kcibul, line 67: nice, I didn't know about this BASH prefix-removal.

kcibul (approved):

.github/workflows/release_workflow.yml
fedorgrab: It was my initial plan, but then I realized that we would have separate runs in the GitHub action tab. Also, I couldn't find a way how to it depended on unit tests. In the current implementation, docker images wouldn't be built if the tests failed.

fedorgrab:
* I have created a composite action. Now unit tests and docker image updates are separate tasks. Didn't find any other way how to implement this, followed [this tutorial](https://wallis.dev/blog/composite-github-actions)
* Now docker images are being updated only on tagged commits on the main branch

kcibul (approved):
lgtm

PR #54, opened 2023-01-25
Progress towards DistributedAnnCollectionSampler

kcibul:
* modified DistributedAnnCollection to be pickle-able (required for DDP)
* added simple DistributedAnnCollectionDataset
* initial DistributedAnnCollectionSampler which does chunk-based sampling
* example script to run with basic AutoEncoder via pytorch lightning

### DIFF
--- a/src/casp/ml/data/distributed_anndata.py
+++ b/src/casp/ml/data/distributed_anndata.py
@@ -84,8 +84,6 @@ class LazyAnnData:
 
     @property
     def cached(self) -> bool:
-        # print(f"filename: {self.filename}")
-        # print(f"cache: {self.cache}")
         return self.filename in self.cache
 
     @property
--- a/src/casp/ml/data/distributed_anndata.py
+++ b/src/casp/ml/data/distributed_anndata.py
@@ -189,8 +189,8 @@ class DistributedAnnCollection(AnnCollection):
 
     def __getstate__(self):
         attributes = self.__dict__.copy()
-        del attributes['cache']
-        del attributes['adatas']        
+        del attributes["cache"]
+        del attributes["adatas"]
         return attributes
 
     def __setstate__(self, d):
--- a/src/casp/ml/data/distributed_anndata.py
+++ b/src/casp/ml/data/distributed_anndata.py
@@ -84,6 +84,8 @@ class LazyAnnData:
 
     @property
     def cached(self) -> bool:
+        # print(f"filename: {self.filename}")
+        # print(f"cache: {self.cache}")
         return self.filename in self.cache
 
     @property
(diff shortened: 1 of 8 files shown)


src/casp/ml/data/distributed_anndata.py
```diff
@@ -185,6 +187,19 @@ def __init__(
         )
         self.var = self.adatas[0].var.copy()
 
+    def __getstate__(self):
+        attributes = self.__dict__.copy()
+        del attributes['cache']
+        del attributes['adatas']        
+        return attributes
+
+    def __setstate__(self, d):
```
fedorgrab, line 196: Is this for pickling? Just curious about the methods.

kcibul: yep -- these are the methods to pickle (__getstate__) and un-pickle (__setstate__)

src/casp/ml/data/distributed_anndata_dataset.py
```diff
@@ -0,0 +1,26 @@
...
+    def __len__(self) -> int:
+        return len(self.dac)
+
+    def __getitem__(self, index: int) -> t.Tuple[Tensor, int]:
+        """
+        :return: Tuple of tensor with a cell gene counts and db index.
+        """
+
+        # TODO: review with yerdos
+        v = self.dac[index]
```
kcibul, line 22: @ordabayevy @fedorgrab -- this is the datatype that was returned in for the original PCA model, just want to verify that this is what we want here.

src/casp/ml/data/distributed_anndata_sampler.py
```diff
@@ -0,0 +1,80 @@
...
+            )
+
+        self.dataset = dataset
+        self.shard_size = shard_size
+        self.seed = seed
+        self.shuffle = shuffle
+        self.num_replicas = num_replicas
+        self.rank = rank
+
+        # TODO: several ways epoch are set, decide on one and use it
```
kcibul, line 41: Pytorch/Lightning can recreate the dataloaders at each epoch, in this is fine.  However, if you don't do that it can call the `set_epoch` method… the "problem" with that is all the work done here in the initializer that depends on the epoch should be moved into `iter()`.  

I think I prefer the latter since it works in either scenario, but would like to discuss

src/example_ddp_anndata.py
```diff
@@ -0,0 +1,151 @@
+import random
```
kcibul, line 1: Is this useful anymore if we are no longer considering an iterable-style dataset?  If Yes -> move these individual classes into there right place in the package hierarchy, if not… vaporize

Issue #58, opened 2023-02-03
Automate BQ operations


fedorgrab:
* extract minibatch to anndata
* convert anndata to avro
* load dataset
* prepare extract

PR #59, opened 2023-02-03
Feat: #58 Automate extract from bq process

fedorgrab:


src/casp/cromwell/bq_ops/extract/extract_command.wdl
```diff
@@ -0,0 +1,59 @@
...
+    String output_bucket_directory
+
+
+    meta {
+        description: "Extract Data from BigQuery to anndata files"
+    }
+
+    scatter(bin_border in bin_borders) {
+        call extract {
+            input:
```
sjfleming, line 49: The neat trick is this... if you have inputs which can just be "passed through" from the workflow-input level to the task-input level, then you don't need to pass them from the workflow to the task.  So here you could just have
```wdl
input:
    start_bin = bin_border[0],
    end_bin = bin_border[1],
```
and leave out everything else.  What will happen then is that the other inputs need to be defined for the task itself, not for the workflow, so (if I'm not mistaken) you'd have to change your input JSON a little bit.  But in my opinion, sometimes this is better, since there's no chance of some kind of unintentional typo.

I think the input JSON would then look like this:
```json
{
   "CASExtractBQ.extract.project_id": "dsp-cell-annotation-service",
   "CASExtractBQ.extract.bq_dataset": "cas_benchmark_v1",
   "CASExtractBQ.extract.extract_table_prefix": "kc_bench_v1",
   "CASExtractBQ.bin_borders": [[0, 9], [10, 19], [20, 29], [30, 39], [40, 49], [50, 59], [60, 69], [70, 79], [80, 89], [90, 99]],
   "CASExtractBQ.extract.output_bucket_name": "dsp-cell-annotation-service",
   "CASExtractBQ.extract.output_bucket_directory": "cromwell_extract_test"
 }
```
so that the task name `extract` is specified to tell the input where it's going (for all except `bin_borders`).

src/casp/cromwell/bq_ops/extract/inputs_extract.json
```diff
@@ -0,0 +1,8 @@
+{
+  "CASExtractBQ.project_id": "dsp-cell-annotation-service",
```
sjfleming, line 2: See comment above about using `CASExtractBQ.extract. ... ` if you want to not have to pass inputs from the workflow to the task

sjfleming (commented):
A little suggestion for how to avoid passing inputs from the workflow to the task, if you want to.  But the way you're doing it is fine, and so people prefer to do it the way you're doing it.

src/casp/services/bq_ops/extract/main.py
```diff
@@ -0,0 +1,85 @@
...
+    end_bin: int,
+    output_bucket_name: str,
+    output_bucket_directory: str,
+) -> None:
+    # It is recommended to use one worker per each CPU core
+    num_of_workers = multiprocessing.cpu_count()
+    with concurrency.ProcessPoolExecutor(max_workers=num_of_workers) as executor:
+        futures = []
+        for bin_number in range(start_bin, end_bin + 1, 1):
+            file_name = f"extract_{bin_number}.hd5ad"
```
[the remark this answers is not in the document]
fedorgrab: Omg, thanks😅

src/casp/cromwell/bq_ops/extract/extract_command.wdl
```diff
@@ -0,0 +1,59 @@
...
+        --dataset ${bq_dataset} \
+        --extract_table_prefix ${extract_table_prefix} \
+        --start_bin ${start_bin} \
+        --end_bin ${end_bin} \
+        --output_bucket_name ${output_bucket_name} \
+        --output_bucket_directory ${output_bucket_directory}
+    }
+
+    runtime {
+        docker: "us-central1-docker.pkg.dev/dsp-cell-annotation-service/cas-services-cicd/cas-pytorch:fg-extract-01"
```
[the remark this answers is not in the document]
fedorgrab: That's a very good comment. I will create a separate issue to include a docker image without a PyTorhc for such cases.

src/casp/cromwell/bq_ops/extract/extract_command.wdl
fedorgrab: That really works, thanks!

sjfleming (approved):
I lack enough familiarity with the codebase to review most of what you've done... but the WDL looks good to me!  :)

src/casp/bq_scripts/extract_minibatch_to_anndata.py
```diff
@@ -116,12 +116,15 @@ def assign_uns_metadata(anndata, json_string):
         anndata.uns[key] = val
 
 
-def extract_minibatch_to_anndata(project, dataset, extract_table_prefix, start_bin, end_bin, output):
+def extract_minibatch_to_anndata(project, dataset, extract_table_prefix, start_bin, end_bin, output, credentials=None):
```
kcibul, line 119: nice addition

src/casp/cromwell/bq_ops/extract/extract_command.wdl
```diff
@@ -0,0 +1,48 @@
...
+        bootDiskSizeGb: 20
+        memory: "20G"
+        cpu: 10
+        zones: "us-central1-a"
+        maxRetries: 0
+        preemptible_tries: 0
```
kcibul, line 30: For cost savings, I'd set this to > 0 (maybe 3)

src/casp/cromwell/bq_ops/extract/inputs_extract.json
```diff
@@ -0,0 +1,8 @@
+{
+  "CASExtractBQ.extract.project_id": "dsp-cell-annotation-service",
+  "CASExtractBQ.extract.bq_dataset": "cas_benchmark_v1",
+  "CASExtractBQ.extract.extract_table_prefix": "kc_bench_v1",
+  "CASExtractBQ.bin_borders": [[0, 9], [10, 19], [20, 29], [30, 39], [40, 49], [50, 59], [60, 69], [70, 79], [80, 89], [90, 99]],
```
kcibul, line 5: It might be nice to have a "bin batch size" of 10, rather than specify the borders here.  It would be easier to optimize your bin sizes, and also not to forget bins (e.g. here there are really 399 bins I think but you're only extracting the first 100)

kcibul (approved):
Looks great!  Just some small comments

Issue #110, opened 2023-11-28
Aggregate (or delete) write requests to DB


fedorgrab:
Instead of writing cas stats for each of the chunks, do it for the whole dataset after the client receives the response.

PR #111, opened 2023-11-28
DB connection pool size increase

fedorgrab:
To allow larger throughput in the system, we need a larger DB connection Pool size. This is the current database connection overload on large datasets.
image
Possible solutions to increase throughput:
1. Implement caching (good for large-scale apps, CAS doesn't look like the one so far).
2. Aggregate write requests. A new issue for this (#110)
3. Increasing database SQL server machine requirements (number of CPUs, RAM). Already did it. In the cloud
4. Increase Database [connection pool](https://en.wikipedia.org/wiki/Connection_pool) size
It seems that neither RAM or CPUs are utilized well, while the number of maximum connections is reaching its limits which (imho) means that the database could easily afford an increase in connection pool size.
image


mbabadi (approved):
I am always a fan of refactoring constants, nice! :-)

PR #116, opened 2023-12-12
Jan Release PR

fedorgrab:


fedorgrab:
I have restructured the microservices. Now, the API and model inference adhere to the pattern of service-oriented microservices. This PR implies a separation of all client communication, [domain logic](https://en.wikipedia.org/wiki/Business_logic), and [data accessing logic](https://en.wikipedia.org/wiki/Data_access_object) from the controller.

The three main components are:
1. Clients: These are communicators responsible for interaction with other services (e.g., MatchingEngineClient, ModelInferenceClient).
2. Data Manager: This software layer is inspired by the [Data Access Object (DAO)](https://en.wikipedia.org/wiki/Data_access_object) pattern. However, this implementation differs slightly from the traditional DAO. DAOs are commonly used in Java development, typically associated with RESTful frameworks and data model CRUD operations, as seen in common examples online. Since our application does not strictly adhere to the RESTful API architecture, I have named this layer the Data Manager layer. It is primarily responsible for any data operation from any type of database used in our project.
3. Service Layer: This layer handles Domain Logic, which refers to the core functions of the application. It utilizes both the Client and Data Manager layers as needed. For example, in the context of annotation, the service layer would incorporate a client request to the model to obtain embeddings. After accumulating them, it would send them for vector search. Once the results are retrieved, the layer would provide match annotations and return them to the controller.
4. Controller (routers): This layer, accessible to users, functions as the interface for interaction with the system.

There are layers of varying depths: clients and the data access manager constitute the deepest layers, the service represents a shallower layer, while routers are the shallowest. The deeper layer should not have any insight into what's happening in the upper layer, whereas the upper layer is aware that it operates with a deeper layer. The only way a deeper layer can communicate with an upper layer is through error throwing. Error throwing is not considered an exception to this rule, because even when the deeper layer throws an error, it still does not know how the error will be handled or by whom (or more precisely, by what entity).

The links where people using FastAPI with patterns. It is all simillar, however not exactly the same as what you would see in this PR. I consider service layer, data layer, client layer and routes layer as the most suitable framework for our microservices.
https://python.plainenglish.io/fastapi-microservice-patterns-domain-driven-design-e99f6f475691
https://medium.com/@yashika51/write-robust-apis-in-python-with-three-layer-architecture-fastapi-and-pydantic-models-3ef20940869c

src/casp/services/api/data_manager/sql_templates/cell_analysis/match_metadata.sql.mako
```diff
@@ -0,0 +1,13 @@
+select
```
mbabadi, line 1: Why is it called match metadata? Shouldn't be called `get_neighborhood_distance_summary` or something like that?

src/casp/services/api/data_manager/sql_templates/cell_analysis/get_cells_by_ids.sql.mako
```diff
@@ -0,0 +1,6 @@
+<%
```
mbabadi, line 1: Let's rename `sql_templates/cell_analysis` to something more descriptive. What's the commonality between all the templates you have (and you intend to have in the future) in this directory?

src/casp/services/api/data_manager/cell_analysis.py
```diff
@@ -0,0 +1,137 @@
...
+        sql_template_data = sql.TemplateData(
+            project=self.project, dataset=cas_model.bq_dataset_name, temp_table_fqn=match_temp_table_fqn
+        )
+        sql_query = sql.render(self.SQL_MATCH_METADATA_DEV_DETAILS, template_data=sql_template_data)
+
+        query_job = self.bigquery_client.query(query=sql_query)
+
+        return bigquery_response_parsers.parse_match_query_job(query_job=query_job, include_dev_details=True)
+
+    def get_cells_by_ids(
```
mbabadi, line 110: `get_cells_by_ids` -> `get_cell_metadata_by_ids`. Also, rename the mako file.

src/casp/services/api/data_manager/cell_analysis.py
```diff
@@ -0,0 +1,137 @@
...
+        sql_template_data = sql.TemplateData(
+            project=self.project, dataset=cas_model.bq_dataset_name, temp_table_fqn=match_temp_table_fqn
+        )
+        sql_query = sql.render(self.SQL_MATCH_METADATA, template_data=sql_template_data)
+
+        query_job = self.bigquery_client.query(query=sql_query)
+
+        return bigquery_response_parsers.parse_match_query_job(query_job=query_job)
+
+    def get_match_query_metadata_dev_details(
```
mbabadi, line 88: `_dev_details` -> `_extended`?

src/casp/services/api/data_manager/cellarium_general.py
```diff
@@ -0,0 +1,110 @@
+import typing as t
+
+from casp.data_manager import BaseDataManager, sql
+from casp.services import settings
+from casp.services.api import schemas
+from casp.services.api.data_manager import exceptions
+from casp.services.db import models
+
+
+class CellariumGeneralDataManager(BaseDataManager):
```
mbabadi, line 10: `CellariumGeneral` -> `CAS`?

src/casp/data_manager/base_data_manager.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+
+    SQL_TEMPLATE_DIR: str
+
+    def __init__(self):
+        credentials, project = utils.get_google_service_credentials()
+        self.credentials = credentials
+        self.project = project
+        self.bigquery_client = bigquery.Client(credentials=self.credentials, project=self.project)
+        self.postgres_db_session = init_db()
```
mbabadi, line 27: How about renaming `postgres_db_session` to `usage_tracking_db_session` or something like that?

src/casp/services/api/schemas/cell_analysis.py
```diff
@@ -0,0 +1,78 @@
...
+    neighbors: t.List[SearchNeighborInfo]
+
+
+class CellariumCellByIdsInput(BaseModel):
+    cas_cell_ids: t.List[int]
+    model_name: str
+    metadata_feature_names: t.List[str]
+
+
+class CellariumCell(BaseModel):
```
mbabadi, line 55: CellariumCell -> CellariumCellMetadata

src/casp/services/api/data_manager/sql_templates/cell_analysis/get_cells_by_ids.sql.mako
mbabadi: Also, could you put documentation at the top of sql.mako templates (e.g. a brief description what the query is supposed to do)?

mbabadi (changes requested):
Nice work! :-) I put in a few comments here and there.

src/casp/data_manager/base_data_manager.py
fedorgrab: The term 'usage_tracking_db_session' doesn't really appeal to me, as we are planning to incorporate more data models into this database. Once we implement cell metadata model in postgres, this name will become outdated, and that change will happen very soon. I believe we should decide on names for these two databases first, which will simplify the process of naming their sessions and clients. Here are my suggestions:

For the PostgreSQL database:

1. 'System Info DB' with the session named 'system_info_db_session'
2. 'System Data DB' with the session named 'system_data_db_session'

For BigQuery, considering that it seems to be used exclusively for storing raw count matrices, we could name it something like 'MatrixDataBase', 'Sparse Matrix DB', or if it is too specific, smth like 'Bigdata DB'?

src/casp/services/api/data_manager/cell_analysis.py
fedorgrab: Well, I thought this feature was supposed to be a tmp solution for 10x for their data analysis. It seems that after the release we would deprecate this feature. Please correct me if I'm wrong.

src/casp/services/api/data_manager/sql_templates/cell_analysis/get_cells_by_ids.sql.mako
mbabadi: How about `sql_templates/cell_operations` or `sql_templates/cell_table_operations`?

mbabadi: To be clear, my issue is with the choice of word `analysis`. To me, analysis means doing a lot more than summarizing some statistics, making temp tables, querying by ID, etc. Operations, on the other hand, seems more relevant.

src/casp/services/api/data_manager/cell_analysis.py
```diff
@@ -0,0 +1,137 @@
...
+from google.cloud import bigquery
+from google.cloud.aiplatform.matching_engine.matching_engine_index_endpoint import MatchNeighbor
+
+from casp.data_manager import BaseDataManager, sql
+from casp.services import settings
+from casp.services.api.data_manager import bigquery_response_parsers, bigquery_schemas, cellarium_general
+from casp.services.db import models
+
+
+class CellAnalysisDataManager(BaseDataManager):
```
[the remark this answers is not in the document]
mbabadi: Likewise, how about `CellOperationsDataManager`?

src/casp/services/api/data_manager/cell_analysis.py
mbabadi: That sounds right, so let's not bother and deprecate in the next release.

src/casp/services/api/data_manager/cell_analysis.py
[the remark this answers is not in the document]
mbabadi: We could also reconsider `CellariumDataStoreDAO` and `CellariumGeneralDAO`?

src/casp/services/api/data_manager/cellarium_general.py
mbabadi: perhaps `CellariumGeneralDAO`?

mbabadi: Please explain to me why `DAO` is inaccurate in our usage?

src/casp/data_manager/base_data_manager.py
mbabadi: Sounds all good to me. For BQ, the most accurate name is `BlockCOOMatrixDatabase`.

src/casp/services/api/data_manager/cellarium_general.py
fedorgrab: DAO stands for Data Access Object. Regarding the terminology, it can be used as an alternative term for Data Manager in our case. Therefore, it's important to note that we should use either DAO or Data Manager, but not both together.
DAO is a programming pattern commonly used in Java development. The current Data Manager is a solution inspired by the DAO pattern, but it doesn't strictly adhere to it, primarily because I, believe, DAO was invented during a time when data and application complexity were more primitive, and people only needed operations for creating, retrieving, updating, and destroying (CRUD) models. 
In our case, the Data Manager is a more complex application layer that enables us to perform complex filtering operations and create temporary tables in a Big Data database. It is similar to DAO in a way that it segregates data operations into a distinct layer, but it also differs from classic DAO. So, to sum up, if we call it `DAO` then it might be confusing because it is not a clear usage of DAO pattern. Just to avoid this confusion, I propose using `DataManager`.

src/casp/services/api/data_manager/cell_analysis.py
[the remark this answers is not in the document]
fedorgrab: Cell Operations sounds like a very good wording to me!

mbabadi (approved):
Looking all good, thank you!! :)

PR #121, opened 2024-02-12
Cellarium ml integration

fedorgrab:


src/casp/services/api/services/cell_operations_service.py
```diff
@@ -81,6 +81,8 @@ def __validate_knn_response(
                 f"Number of query ids ({len(embeddings)}) and knn matches ({len(knn_response)}) does not match. "
                 f"This could probably be caused by Vector Search overload."
             )
+        if len(knn_response[0]) == 0:
+            raise exceptions.VectorSearchResponseError("Vector Search returned match with 0 neighbors.")
```
ordabayevy, line 85: nit: `a match`

src/casp/services/deploy/Dockerfile.pytorch
```diff
@@ -4,16 +4,18 @@ WORKDIR /app
...
 
 # Install Dependencies
 COPY requirements.txt .
 COPY test-requirements.txt .
-RUN pip install Cython numpy
+#RUN pip install Cython numpy
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
-RUN pip install -r test-requirements.txt
+#RUN pip install -r test-requirements.txt
```
ordabayevy, line 18: Were you going to remove these lines or keep for a future reference?

src/casp/services/model_inference/services.py
```diff
@@ -1,13 +1,14 @@
-import pickle
 import typing as t
 
 import anndata
-import torch
-from google.cloud import storage
+import numpy
+import numpy as np
```
ordabayevy, line 5: Double import

ordabayevy (changes requested):
Looks great, just a few comments.

src/casp/services/deploy/Dockerfile.pytorch
fedorgrab: Oh right, thanks!

src/casp/services/_settings.py
```diff
@@ -87,6 +87,9 @@ class ProductionSettings(AllEnvSettings):
 class LocalSettings(AllEnvSettings):
     # General
     debug = True
+    MODEL_SERVER_URL: str = "http://localhost:8001"
+    API_SERVICE_PORT: int = 8000
+    MODEL_SERVICE_PORT: int = 8001
```
KevinCLydon, line 92: This is maybe a silly question, but why is there are `MODEL_SERVICE_PORT` variable if the port is already set in `MODEL_SERVER_URL`?  Do they serve different purposes?

src/casp/services/model_inference/services.py
```diff
@@ -18,34 +19,81 @@ def __init__(self):
...
+        :raises ModelOutputError: If the length of obs_ids and embeddings are not the same or if the number of embedding
+            dimensions is not equal to the model's embedding dimensions.
         """
-        adata = anndata.read_h5ad(file)
-        X = torch.Tensor(adata.X.todense().astype(int))
-        obs_ids = adata.obs.index.values.astype(str).tolist()
-        return X, obs_ids
+        if embeddings.shape[0] != len(obs_ids):
+            raise exceptions.ModelOutputError(
+                "Model Output Error: Length of obs_ids and embeddings should be the same."
```
KevinCLydon, line 89: Would it be valuable to include the lengths for `obs_ids` and `embeddings` in this message?

src/casp/services/_settings.py
fedorgrab: Oh, this is because `MODEL_SERVER_URL` is used by api manager to access it, while `MODEL_SERVICE_PORT` is used by gunicorn to serve the model service app. I agree, it is a bit confusing tho.

src/casp/services/model_inference/services.py
fedorgrab: Absolutely, yes! I will add it.

fedorgrab:
Thanks for the review @KevinCLydon @ordabayevy I pushed new changes, please have a look. Also, @KevinCLydon regarding your comment:
https://github.com/cellarium-ai/cellarium-cloud/pull/121#discussion_r1488618495
do you think I should make more clear in the code and reconsider this approach?

KevinCLydon:
> Also, @KevinCLydon regarding your comment: [#121 (comment)](https://github.com/cellarium-ai/cellarium-cloud/pull/121#discussion_r1488618495) do you think I should make it more clear in the code and reconsider this approach?

I think it's probably fine as it is.  If you want it to be extra clear, you could maybe add a couple quick comments next to them to explain what they're used for, but I don't think it's totally necessary.

ordabayevy (approved):

KevinCLydon (approved):

PR #126, opened 2024-02-16
Chore: Extract: Paths, filter exclude 

fedorgrab:
This PR includes:

1. Changing extract paths for "extract" and "prepare extract" scripts, along with updated examples.
2. Adding exclusions in the filter with two additional filter clauses: __not_in and __not_eq.
3. Fixing the SQL rendered query that had issues with the ' (apostrophe) in its filter. When I provided an assay type containing an apostrophe, it was breaking the SQL query because each string was wrapped in single quotes.


fedorgrab:
This is redundant and will be addressed in a benchmarking PR.

PR #134, opened 2024-03-05
Reduced matching engine request batch size and added retry logic

KevinCLydon:
Match requests to matching engine will now be broken into chunks of 5 (determined by `GET_MATCHES_CHUNK_SIZE`) to be submitted sequentially.  Also added retry logic using the tenacity library.  The retry scheme right now uses 4 attempts per request, with exponential backoff for the wait time between requests.


src/casp/services/api/services/cell_operations_service.py
```diff
@@ -17,6 +19,8 @@
 
 AVAILABLE_FIELDS_DICT = set(schemas.CellariumCellMetadata.__fields__.keys())
 
+GET_MATCHES_CHUNK_SIZE = 5
```
nmalfroy, line 22: Should we make this configurable?

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -95,23 +99,90 @@ def get_knn_matches(self, embeddings: np.array, model_name: str) -> t.List[t.Lis
...
         )
-        self.__validate_knn_response(embeddings=embeddings, knn_response=matches)
+        self.__validate_knn_response(embeddings=embeddings_chunk, knn_response=matches)
         return matches
 
     def get_knn_matches_as_dict(self, embeddings: np.array, model_name: str) -> t.List[t.List[t.Dict[str, t.Any]]]:
         index, index_endpoint_client = self.__get_match_index_endpoint_client_for_model(model_name=model_name)
 
+        # Break embeddings into chunks so we don't overload the matching engine
+        num_chunks: int = math.ceil(len(embeddings) / GET_MATCHES_CHUNK_SIZE)
```
nmalfroy, line 153: Should we pull this into it's own method since it gets used in a few places.  Also maybe worth adding a unit test for said method?

nmalfroy (approved):
This looks great!  tenacity lib seems super useful.  A couple comments. I don't think they're blockers to merge but might be worth doing?  Your call

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: That's a good idea.  I could move it to the settings file and make it optionally configurable with an environment variable maybe?  I think that's the convention we're using currently.  Is that right @fedorgrab ?

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: Those are both good ideas.  I can make those changes

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -95,23 +99,90 @@ def get_knn_matches(self, embeddings: np.array, model_name: str) -> t.List[t.Lis
...
+        embeddings_chunks = np.array_split(embeddings, num_chunks)
+
+        all_matches = []
+        for i in range(0, num_chunks):
```
fedorgrab, line 157: Seems like this loop is identical to the one that's in `get_knn_matches` except for the method that's used to called inside of the retryer. 
Do you guys think it would be a good idea to wrap it up in some kind of a method that takes callback and its arguments (and maybe num_attempts)?

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -95,23 +99,90 @@ def get_knn_matches(self, embeddings: np.array, model_name: str) -> t.List[t.Lis
...
+            # Set up retry logic for for the matching engine requests
+            retryer = Retrying(
+                stop=stop_after_attempt(4),
+                wait=wait_exponential(multiplier=2, min=0, max=30) + wait_random(0, 2),
```
fedorgrab, line 161: I think making constants (or setting constants) from these numbers might be also a good idea).

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -95,23 +99,90 @@ def get_knn_matches(self, embeddings: np.array, model_name: str) -> t.List[t.Lis
...
 
+        # Break embeddings into chunks so we don't overload the matching engine
+        num_chunks: int = math.ceil(len(embeddings) / GET_MATCHES_CHUNK_SIZE)
+        embeddings_chunks = np.array_split(embeddings, num_chunks)
+
+        all_matches = []
+        for i in range(0, num_chunks):
+            # Set up retry logic for for the matching engine requests
+            retryer = Retrying(
+                stop=stop_after_attempt(4),
```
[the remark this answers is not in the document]
fedorgrab: Also, I think we should increase number of attempts to a higher number, maybe like at least 10.

fedorgrab (changes requested):
This is awesome! I did a stress test and the index showed up to 87 qps with almost no errors. 

I've left some small comments.

Also, could you please add the changes into the CHANGELOG.rst under existing 1.4.2 version? [Here's an example](https://github.com/cellarium-ai/cellarium-cas/blob/development/CHANGELOG.rst) of what sections can have the changelog file

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: Changes made!

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: Okay cool.  I added it to the settings file

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -95,23 +99,90 @@ def get_knn_matches(self, embeddings: np.array, model_name: str) -> t.List[t.Lis
...
+                wait=wait_exponential(multiplier=2, min=0, max=30) + wait_random(0, 2),
```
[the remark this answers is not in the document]
KevinCLydon: It's super convenient

src/casp/services/api/services/cell_operations_service.py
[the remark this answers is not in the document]
KevinCLydon: Moved it to settings and set the default to 10

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: I consolidated it into one function as you suggestion.  I changed two existing functions to wrappers that just pass the correct callback

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: Moved them all to settings

fedorgrab (approved):
Small cosmetic things, but please add updates to changelog before merging.

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -93,27 +137,72 @@ def get_knn_matches(self, embeddings: np.array, model_name: str) -> t.List[t.Lis
...
+            model_name=model_name,
+            chunk_matches_function=self.__get_knn_matches_as_dict_for_chunk,
+        )
 
+    def __get_knn_matches_as_dict_for_chunk(
+        self,
+        embeddings_chunk: np.array,
+        index: models.CASMatchingEngineIndex,
+        index_endpoint_client: clients.CustomMatchingEngineIndexEndpointClient,
+    ) -> t.List[t.List[MatchNeighbor]]:
```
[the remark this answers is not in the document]
KevinCLydon: Thanks for catching this one too

KevinCLydon:
> Small cosmetic things, but please add updates to changelog before merging.

@fedorgrab The previous entry in the changelog has a header with version number 1.4.2.  Should I use a new version number?

fedorgrab:
> > Small cosmetic things, but please add updates to changelog before merging.
> 
> @fedorgrab The previous entry in the changelog has a header with version number 1.4.2. Should I use a new version number?

No, let's keep using 1.4.2, we well start a new version after the release. Version increments after each merge dev -> main, while dev aggregates all changelogs from feature branches.

fedorgrab:
> > > > Small cosmetic things, but please add updates to changelog before merging.
> > > 
> > > 
> > > @fedorgrab The previous entry in the changelog has a header with version number 1.4.2. Should I use a new version number?
> > 
> > 
> > No, let's keep using 1.4.2, we well start a new version after the release. Version increments after each merge dev -> main (which is basically a release), while dev aggregates all changelogs from feature branches.
> 
> @fedorgrab Okay I've updated it. Can you take a quick look to make sure I did it correctly?

Looks Great!

fedorgrab (approved):

nmalfroy (approved):
looks great!

PR #141, opened 2024-03-19
REST API index support

nmalfroy:
Sorry, this PR got a little big but I think it does some good stuff.  I tried to organize commit-by-commit so that should make a little more consumable.

Vertex AI Vector search allows you to configure an index to be hosted using the a REST (as opposed to gRPC) approach.  This is a lot easier to debug and interact with but might be a little slower so we want to give users the option to do both.  Because both use fairly different APIs, the interaction was wrapped in a common interface with a common `MatchResult` class.  With this PR we will have the choice of which endpoint method to use per vector search index.

So in more detail, what does this PR do:
- It makes sure that migrations are sorted alphabetically.  E.g. there are two migrations in this PR and the files end up looking like:
image
- Adds support for REST API fronted Vertex search indexes (there are 2 new fields on the index table that can be set in the admin UI too:
   - is grpc
   - api endpoint (needed for the REST API)
- Introduces mockito and parameterized libraries to unit test these calls more easily
- Removed a unique constraint on the model path field since we may want the same model to be used by different index objects (note: we could have kept the constraint and just changed the index to allow sourcing from the same model but that would have gotten a lot trickier)
- In order to support duplicating objects, added a new "Clone" action in the admin view which allows admins to create new users, models and indexes with the same initial values as existing objects)
image
- Added instructions for testing gRPC index queries locally (it does require being on the VPN but is really helpful)
- Drive by: made the `query-cells-by-id` return the valid fields to use if an invalid field was passed in

General things:
- code coverage gets run on `tox -e unit` now
- git ignore the virtual env and vscode config (we may want to check that in but happy to hold off)


src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,168 @@
...
+        id: str
+        distance: float
+        feature_vector: List[float] = None
+
+    class NearestNeighbors(NamedTuple):
+        """
+        NearestNeighbors represents the result of a nearest neighbors request to the matching service. There should be 1
+        per query that was sent.
+        """
+        neighbors: List["MatchResult.Neighbor"] = []
```
nmalfroy, line 31: double check if quotes are needed

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
+from typing import List, NamedTuple
+
+from google.cloud import aiplatform_v1
+from google.cloud.aiplatform.matching_engine.matching_engine_index_endpoint import MatchNeighbor
+
+from casp.services.api import clients
+from casp.services.db import models
+
+"""
```
fedorgrab, line 9: Style thing: Module level docstrings usually go before imports.

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+
+        :return: The new client.
+        """
+        if index.is_grpc:
+            return MatchingClientGRPC(index=index)
+        else:
+            return MatchingClientREST(index=index)
+
+
+class MatchingClientGRPC(MatchingClient):
```
KevinCLydon, line 79: This and MatchingClientREST could maybe each benefit from a short comment explaining how they differ and why they both exist, as opposed to just having the one client.

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -68,28 +72,24 @@ async def get_embeddings(file_to_embed: t.BinaryIO, model_name: str) -> t.Tuple[
...
             raise exceptions.VectorSearchResponseError(
-                f"Number of query ids ({len(embeddings)}) and knn matches ({len(knn_response)}) does not match. "
+                f"Number of query ids ({len(embeddings)}) and knn matches ({len(knn_response.matches)}) does not match. "
                 f"This could probably be caused by Vector Search overload."
             )
-        if len(knn_response[0]) == 0:
-            raise exceptions.VectorSearchResponseError("Vector Search returned a match with 0 neighbors.")
 
-    def __get_knn_matches_with_retry(
+        for match in knn_response.matches:
```
KevinCLydon, line 86: Is this a possible error mode we weren't accounting for before, or is this something specific to the switch to using MatchResult?

src/casp/services/db/alembic.ini
```diff
@@ -8,7 +8,7 @@ script_location = casp/services/db/migrations
 # Uncomment the line below if you want the files to be prepended with date and time
 # see https://alembic.sqlalchemy.org/en/latest/tutorial.html#editing-the-ini-file
 # for all available tokens
-# file_template = %%(year)d_%%(month).2d_%%(day).2d_%%(hour).2d%%(minute).2d-%%(rev)s_%%(slug)s
+file_template = t_%%(year)4d-%%(month).2d-%%(day).2d_%%(hour).2d-%%(minute).2d-%%(second).2d_%%(slug)s
```
KevinCLydon, line 11: Big fan of this change

src/casp/services/db/migrations/versions/t_2024-03-13_14-58-14_non_unique_model_path.py
```diff
@@ -0,0 +1,27 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "a2e0ae902a06"
+down_revision = "2633d266e26d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
KevinCLydon, line 19: Same here.  We can probably get rid of these comments.  Unless we want to remember which changes were autogenerated?

src/casp/services/db/migrations/versions/t_2024-03-13_14-58-14_non_unique_model_path.py
```diff
@@ -0,0 +1,27 @@
...
+    op.drop_constraint("cas_model_model_file_path_key", "cas_model", type_="unique")
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint("cas_model_model_file_path_key", "cas_model", ["model_file_path"])
```
KevinCLydon, line 26: This could potentially fail if there is data within the table when we try to downgrade that violates this constraint.  Do we want to have some way of handling that here, or would it better to let it fail so whoever is running this knows they have to manually make some data changes?

src/casp/services/db/models.py
```diff
@@ -24,7 +24,7 @@ def __repr__(self):
 class CASModel(db.Base):
```
KevinCLydon, line 24: Maybe outside the scope of this PR and something I'd also want @fedorgrab to weigh in on: These classes are used in several places in the project and aren't documented at all.  It's admittedly not that difficult to understand what they are, but someone coming to this fresh (or returning after a while) might be confused about at least some of the columns, like `model_file_path`.  I think this file would probably benefit from a quick documentation pass.

tests/unit/test_cell_operations_service.py
```diff
@@ -2,28 +2,301 @@
...
+        self,
+        model: models.CASModel = MODEL,
+        index: models.CASMatchingEngineIndex = INDEX,
+        anndata_data: bytes = ANNDATA_DATA,
+        embeddings: List[List[float]] = [],
+        matching_client_response: Optional[MatchResult] = None,
+    ) -> MatchResult:
+        """
 
+        Mock call to the model embeding service and the matching client.
```
KevinCLydon, line 253: Nitpick: "embedding"

tests/unit/test_cell_operations_service.py
```diff
@@ -2,28 +2,301 @@
...
+        )
+
+        # mock calls to get cell distribution
+        temp_table_fqn = "temp_table_fqn"
+        query_ids = [f"q{i}" for i in range(len(embeddings))]
+        response = [
+            {"query_cell_id": query_ids[i], "matches": [{"cell_type": cell_types[i], "cell_count": 10}]}
+            for i in range(len(embeddings))
+        ]
+        when(self.cell_operations_service.cell_operations_dm).insert_matches_to_temp_table(
```
KevinCLydon, line 148: I admittedly haven't used mockito in a while, so I might be misremembering and worrying over nothing.  What happens here if `insert_matches_to_temp_table` is called with any other inputs than the ones you've mocked up a response for?  Does it attempt to call the actual method instead of the mock, or does it still call the mock and then throw an error or something?

tests/unit/test_cell_operations_service.py
```diff
@@ -2,28 +2,301 @@
...
+            model
+        )
+        when(self.cell_operations_service.cellarium_general_dm).get_index_for_model(model_name=model_name).thenReturn(
+            index
+        )
 
-def test_split_embeddings_into_chunks_uneven_split() -> None:
-    embeddings = numpy.random.rand(100, 100)
-    chunk_size = 9
+        # mock calls to that use the matching client
```
KevinCLydon, line 281: I think this is also a typo?

KevinCLydon (changes requested):
This looks good!  I like the new unit test stuff.  It's very organized.  I mostly have minor questions and nitpicky stuff.

src/casp/services/db/models.py
fedorgrab: Yeah, I agree 100%, we can add docstrings in rst format in class description. This will allow us to autogenerate docs for sphinx documentation page as well. Also, I think db should go out of services and be a part of data_manager boilerplate package in the repo.

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -28,9 +28,13 @@ class CellOperationsService:
...
 
-    def __init__(self):
-        self.cell_operations_dm = CellOperationsDataManager()
-        self.cellarium_general_dm = CellariumGeneralDataManager()
+    def __init__(
+        self,
+        cell_operations_dm: CellOperationsDataManager = None,
+        cellarium_general_dm: CellariumGeneralDataManager = None,
+    ):
+        self.cell_operations_dm = cell_operations_dm if cell_operations_dm else CellOperationsDataManager()
```
fedorgrab, line 36:
```suggestion
        self.cell_operations_dm = cell_operations_dm or CellOperationsDataManager()
```

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -28,9 +28,13 @@ class CellOperationsService:
...
+        self.cellarium_general_dm = cellarium_general_dm if cellarium_general_dm else CellariumGeneralDataManager()
```
fedorgrab, line 37:
```suggestion
        self.cellarium_general_dm = cellarium_general_dm or CellariumGeneralDataManager()
```

src/casp/services/api/clients/matching/matching_client.py
nmalfroy: ah it is

docs/source/modules/running_locally.rst
```diff
@@ -25,3 +25,36 @@ Most of the methods will reuqire you to be authenticated. To do this, you'd need
 
 
 Once the Admin service is running, you can visit `Admin Dashboard <http://localhost:5000>`_ to create a user and token.
+
+To run commands that require accessing the vector search API using gRPC, you will need to create a tunnel.  To do this you
+must:
+- be on the Broad non-split VPN and have your
```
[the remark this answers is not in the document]
nmalfroy: derp.  No, i went to a new bullet point and forgot to fix the sentence

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+    class NearestNeighbors(NamedTuple):
+        """
+        NearestNeighbors represents the result of a nearest neighbors request to the matching service. There should be 1
+        per query that was sent.
+        """
+
+        neighbors: List["MatchResult.Neighbor"] = []
+
+    """
+    MatchResult represents the result of a match request to the matching service.
```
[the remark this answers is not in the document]
nmalfroy: ah good catch. I had it there and then moved subclasses in and put the docstring in the wrong place

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+        :return: The concatenated match result as a new instance of MatchResult.
+        """
+        return MatchResult(matches=self.matches + other.matches)
+
+
+class MatchingClient:
+    def __init__(self, index: models.CASMatchingEngineIndex) -> None:
+        self.index = index
+
+    async def match(self, queries: List[List[float]]) -> MatchResult:
```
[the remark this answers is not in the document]
nmalfroy: great call.  I just added a `raise NotImplementedError`

src/casp/services/db/migrations/versions/t_2024-03-13_14-50-44_rest_support_for_index.py
```diff
@@ -0,0 +1,40 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "2633d266e26d"
+down_revision = "0009"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
nmalfroy: Yeah, I wasn't sure what we wanted to do there since they're in the other classes.  Happy to remove

src/casp/services/api/clients/matching/matching_client.py
nmalfroy: Totally.  Will do

src/casp/services/api/services/cell_operations_service.py
nmalfroy: Yeah, that's what it looks like.  We were only looking for the first match to not have neighbors.  It seemed like a better idea to make sure every single match had neighbors. 
 @fedorgrab does that seem right?

src/casp/services/db/models.py
nmalfroy: Agreed.  If that's ok, I might wait for a later PR here.  This one's already a bit chonky

tests/unit/test_cell_operations_service.py
nmalfroy: Ah it does not because I mocked the two data manager classed (on line 45).  That's by I made the `cell_operations_dm` and `cellarium_general_dm` constructor parameters with defaults.  That lets the unit test pass in safe mocks that will return `None` if not stubbed out (just confirmed the behavior locally to be sure)

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        Match queries against the specified index.
+
+        :param index: The index to match against.
```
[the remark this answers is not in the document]
nmalfroy: gah good call.  That's left over from my original implementation

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
+from typing import List, NamedTuple
+
+from google.cloud import aiplatform_v1
+from google.cloud.aiplatform.matching_engine.matching_engine_index_endpoint import MatchNeighbor
+
+from casp.services.api import clients
+from casp.services.db import models
+
+"""
+Contains a common interface for querying the matching engine regarless of API (e.g. gRPC vs REST)
```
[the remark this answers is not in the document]
nmalfroy: Thanks for that!  I wasn't sure.  I'm used to Java where there's a file per class (baring subclasses).  I'll combine that with @KevinCLydon 's comment and just put that as part of the `MatchResult`'s doc string

src/casp/services/api/services/cell_operations_service.py
fedorgrab: I've never seen such cases, that if there was only one missing (basically I saw only results where either all of them were empty arrays, or none).

However, being extra cautious doesn't seem like a bad thing to me, especially if it doesn't increase the runtime, which I believe is the case.

Also, if we're going there, might be a good idea to check if the length of the nearest neighbors array equals to the embedding dimension, not just comparing it with a zero.

src/casp/services/db/models.py
fedorgrab: Totally, this could be like a separate PR itself.:)

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+        :return: The new client.
+        """
+        if index.is_grpc:
+            return MatchingClientGRPC(index=index)
+        else:
+            return MatchingClientREST(index=index)
+
+
+class MatchingClientGRPC(MatchingClient):
+    def __init__(self, index: models.CASMatchingEngineIndex) -> None:
```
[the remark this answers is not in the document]
nmalfroy: Ah sadly that's not true.  That only works for empty constructors (same in Java)

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+            num_neighbors=self.index.num_neighbors,
+        )
+
+        return self.__adapt_result(matches)
+
+    def __get_match_index_endpoint_client(self) -> clients.CustomMatchingEngineIndexEndpointClient:
+        return clients.CustomMatchingEngineIndexEndpointClient(index_endpoint_name=self.index.endpoint_id)
+
+    def __adapt_result(self, result: List[List[MatchNeighbor]]) -> MatchResult:
+        return MatchResult(
```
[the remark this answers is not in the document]
nmalfroy: absolutely.  That's why I'd asked originally.  I was sort of on the fence.  More than happy to change this (I find Python's lambda looping syntax a little hard to read...might just be habit though)

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+                        )
+                    ),
+                    result,
+                )
+            )
+        )
+
+
+class MatchingClientREST(MatchingClient):
+    def __init__(self, index: models.CASMatchingEngineIndex) -> None:
```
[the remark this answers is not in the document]
nmalfroy: yeah sadly I have to leave that

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+
+from casp.services.api import clients
+from casp.services.db import models
+
+"""
+Contains a common interface for querying the matching engine regarless of API (e.g. gRPC vs REST)
+"""
+
+
+class MatchResult(NamedTuple):
```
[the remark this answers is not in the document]
nmalfroy: I'm trying to get that to work but Pydantic seems really unhappy about nested classes

tests/unit/test_cell_operations_service.py
```diff
@@ -2,28 +2,301 @@
...
+    is_grpc=True,
+    endpoint_id="endpoint_id_grpc",
+    deployed_index_id="deployed_index_id_grpc",
+    num_neighbors=3,
+    model_id=1,
+)
+ANNDATA_DATA = b"testdata"
+
+
+class TestCellOperationsService:
```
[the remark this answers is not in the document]
nmalfroy: I like not having to explicitly call the setup and teardown methods for each test method

src/casp/services/api/clients/matching/matching_client.py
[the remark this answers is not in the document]
nmalfroy: Also FWIW, I left `MatchClient` as a `NamedTuple` since it's not technically an object model

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+        :param index: The index to match against.
+        :param queries: The queries to match.
+
+        :return: The match result.
+        """
+
+        vector_search_client = self.__get_match_index_endpoint_client()
+
+        # Build query objects
+        query_objects = list(
```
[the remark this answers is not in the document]
nmalfroy: fair enough!

src/casp/services/db/README.rst
```diff
@@ -79,6 +79,8 @@ Database Migrations
 
 `Alembic <https://alembic.sqlalchemy.org/en/latest/>`_ is used for managing database migrations.
 
+Note: migration operations must be run from the cellarium-cloud/src directory.
```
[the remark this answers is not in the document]
nmalfroy: ooh, cool! will update

src/casp/services/db/alembic.ini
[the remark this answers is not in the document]
nmalfroy: I changed it slightly from that commented version. I think they just have it there as an example

test-requirements.txt
```diff
@@ -1,6 +1,9 @@
 pytest
+pytest-asyncio
 pytest-cov
 pytest-dependency
+parameterized
```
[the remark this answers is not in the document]
nmalfroy: oh gotcha.  I thought we only did that for the main dependencies.  Happy to add

tests/unit/test_cell_operations_service.py
```diff
@@ -2,28 +2,301 @@
 Tests things in the cell_operations_service that can reasonably be tested with a unit test.
 """
 
+import io
+import re
+from typing import Any, Dict, List, Optional
```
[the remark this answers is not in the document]
nmalfroy: Yeah, I saw that we had it a couple of different ways (I saw it this way in some of the files to.  Happy to switch to make it consistent.  I'll update all modules to use `import as t`

src/casp/services/api/clients/matching/matching_client.py
```diff
@@ -0,0 +1,188 @@
...
+            return MatchingClientGRPC(index=index)
+        else:
+            return MatchingClientREST(index=index)
+
+
+class MatchingClientGRPC(MatchingClient):
+    def __init__(self, index: models.CASMatchingEngineIndex) -> None:
+        super().__init__(index)
+
+    async def match(self, queries: List[List[float]]) -> MatchResult:
```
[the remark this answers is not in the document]
nmalfroy: Gotcha.  I wasn't sure if we wanted the convention of "private" methods being under the public ones (which is what I see codebases generally do) but maybe that's more of a java convention

src/casp/services/api/clients/matching/__init__.py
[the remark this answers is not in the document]
nmalfroy: I was thinking it might make sense to have it in a separate package in case we need to add more.  I'll move CustomClient in there too

src/casp/services/api/services/cell_operations_service.py
nmalfroy: ooh, that's a nice convention!

src/casp/services/api/clients/matching/__init__.py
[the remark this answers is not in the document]
nmalfroy: Eh actually, maybe I'll move it out of the package.  Still getting used to classes vs. modules vs. packages

KevinCLydon (approved):
Looks good to me! 👍

fedorgrab (approved):
Great stuff!

Issue #94, opened 2023-09-12
Great `bq_scripts` refactor


fedorgrab:

fedorgrab:
During PR #101 the following comment is highlighted:
"Unrelated to this PR, but we should never simply catch a broad exception, print the error message, and move on. Let's fix this during the great big refactor"

fedorgrab:
Another comment: make sure to get rid of print statements and establish an appropriate logger.

fedorgrab:
Closing this as after refactor (#206 ) it is not needed anymore, BigQuery is a legacy supported data repository, manage in [Cellarium Nexus](https://github.com/cellarium-ai/cellarium-nexus)

PR #144, opened 2024-04-05
Consensus Engine

fedorgrab:
Changes:
1. Cell Metadata models and querying. Also check out how a temp table query could be executed in SQLalchemy, no raw SQL required, which seems cool. 
2. Admin restructure. Since we now getting more and more models in admin, seems reasonable to group them together. Now admin page nav bar looks like:
image
Where All user related models go under User, all ML realted models go under ML Management and all Cell Metadata models go under Cell Data management.
3. Consensus Engine which includes ontology_aware (new method that uses postgres for cell metadata fetching), and the old method which is now called cell_type_count. Old method should be deprecated and rewritten with use of postgres as well, but not in this PR.:)
4. In local environment SQLAlchemy will print all SQL queries
5. Inspired by @nmalfroy 's PR (#141), adding more pydantic models to wrap internal data flows into the python data classes


src/casp/services/api/data_manager/bigquery_response_parsers.py
```diff
@@ -2,27 +2,29 @@
 
 from google.cloud import bigquery
 
+from casp.services.api import schemas
 
-def _parse_match_object(row: bigquery.Row) -> t.Dict[str, t.Any]:
+
+def _parse_match_object(row: bigquery.Row) -> schemas.AnnotationInfoCellTypeCount:
```
nmalfroy, line 8: We might want to rename the module too if it's not going to be BQ anymore

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -17,15 +19,20 @@ class CellOperationsDataManager(BaseDataManager):
...
+        f"{CELL_OPERATIONS_TEMPLATE_DIR}/get_neighborhood_distance_summary_dev_details.sql.mako"
     )
-    SQL_GET_CELLS_BY_IDS = f"{CELL_ANALYSIS_TEMPLATE_DIR}/get_cell_metadata_by_ids.sql.mako"
+    SQL_GET_CELLS_BY_IDS = f"{CELL_OPERATIONS_TEMPLATE_DIR}/create_temp_table_for_metadata.sql.mako.mako"
+    # SQLAlchemy parsable templates
+    SQL_CREATE_CELL_TEMP_TABLE = f"{CELL_OPERATIONS_TEMPLATE_DIR}/create_temp_table_for_metadata.sql"
+    SQL_GET_CELL_METADATA_FROM_TEMP_TABLE = f"{CELL_OPERATIONS_TEMPLATE_DIR}/get_cell_metadata_from_temp_table.sql"
     redis_cache_key_prefix = "cell_operations"
+    # Settings and constants
+    MAX_CELL_IDS_PER_QUERY = 20_000  # Maximum number of cell IDs that can be queried at once
```
nmalfroy, line 35: should this be configurable?

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -112,52 +119,69 @@ def get_neighborhood_distance_summary_dev_details(
...
-        # If there are missed_ids, retrieve them from the database
-        if missed_ids:
-            columns = [column.key for column in CellInfo.__table__.columns if column.key != "obs_metadata_extra"]
-            with_entity_items = [getattr(CellInfo, name) for name in columns]
-            database_cells = (
-                self.system_data_db_session.query(CellInfo)
-                .with_entities(*with_entity_items)
-                .filter(CellInfo.cas_cell_index.in_(missed_ids))
-                .all()
+        if not cell_ids:
```
nmalfroy, line 132: Ah I see you moved away from the caching :)

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -112,52 +119,69 @@ def get_neighborhood_distance_summary_dev_details(
...
+            raise ValueError("No cell IDs provided")
+
+        if len(cell_ids) > self.MAX_CELL_IDS_PER_QUERY:
+            raise ValueError(f"Number of cell IDs exceeds the maximum of {self.MAX_CELL_IDS_PER_QUERY}")
```
nmalfroy, line 136: Not for this PR but I think we lose this information with the global exception handler.  We should find a way to more reliably return this information to the end users

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -112,52 +119,69 @@ def get_neighborhood_distance_summary_dev_details(
...
+
+        with self.system_data_db_session.begin():
+            # Creating temp table
```
nmalfroy, line 139: should we add a comment to specify that the temp table gets auto vacuumed when session ends?

KevinCLydon (commented):
I have one very minor note/question, but otherwise it looks good to me!

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -170,67 +170,66 @@ def __split_embeddings_into_chunks(embeddings: np.array, chunk_size: int) -> t.L
...
-            )
-
-        return self.cell_operations_dm.get_neighborhood_distance_summary(
-            cas_model=cas_model, match_temp_table_fqn=temp_table_fqn
-        )
-
-    async def annotate_adata_file(
-        self, user: models.User, file: t.BinaryIO, model_name: str, include_dev_metadata: bool
-    ) -> t.List[t.Dict[str, t.Any]]:
+        consensus_strategy: str,
```
nmalfroy, line 178: This should be an enum

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,226 @@
...
+
+        :param query_ids: List of query cell IDs from the kNN query.
+        :param knn_query: The result of the kNN query.
+
+        :return: A list of summarized query neighbor context annotations.
+        """
+        ...
+
+
+class ConsensusStrategyType:
```
nmalfroy, line 27: maybe make this explicitly extend enum
```suggestion
class ConsensusStrategyType(t.Enum):
```

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,226 @@
...
+    :param cell_operations_dm: Cell operations data manager object.
+    """
+
+    REQUIRED_CELL_INFO_FEATURE_NAMES = ["cas_cell_index", "cell_type", "cell_type_ontology_term_id"]
+
+    def __init__(
+        self,
+        normalize: bool = True,
+        prune_threshold: float | None = 0.1,
+        cell_ontology_resource: CellOntologyResource | None = None,
```
nmalfroy, line 130: I thought we used `t.Optional` for this?

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,226 @@
...
+    :param cell_ontology_resource: Cell ontology resource object.
+    :param cell_operations_dm: Cell operations data manager object.
+    """
+
+    REQUIRED_CELL_INFO_FEATURE_NAMES = ["cas_cell_index", "cell_type", "cell_type_ontology_term_id"]
+
+    def __init__(
+        self,
+        normalize: bool = True,
+        prune_threshold: float | None = 0.1,
```
nmalfroy, line 129: It's kinda weird to have this be nullable but the default is not None

src/casp/services/db/__init__.py
```diff
@@ -9,6 +9,7 @@
     max_overflow=settings.DB_CONNECTION_POOL_MAX_OVERFLOW,
     pool_timeout=settings.DB_CONNECTION_POOL_TIMEOUT,
     pool_recycle=settings.DB_CONNECTION_POOL_RECYCLE,
+    echo=settings.ENVIRONMENT == "local",
```
nmalfroy, line 12: neat!

src/casp/services/db/migrations/versions/t_2024-04-05_8-29-42_add_cell_related_models.py
```diff
@@ -10,14 +10,13 @@
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision = "0010"
-down_revision = "0009"
+revision = "86eda43c427a"
```
nmalfroy: Ah nevermind my previous comment about naming!

src/casp/services/db/models/cells.py
```diff
@@ -43,26 +43,26 @@ class CellInfo(db.Base):
...
-    assay = sa.Column(sa.String(255), nullable=True)
-    development_stage = sa.Column(sa.String(255), nullable=True)
-    tissue = sa.Column(sa.String(255), nullable=True)
-    disease = sa.Column(sa.String(255), nullable=True)
-    organism = sa.Column(sa.String(255), nullable=True)
-    self_reported_ethnicity = sa.Column(sa.String(255), nullable=True)
-    sex = sa.Column(sa.String(255), nullable=True)
-    suspension_type = sa.Column(sa.String(255), nullable=True)
-    total_mrna_umis = sa.Column(sa.Integer, nullable=True)
+    cell_type = sa.Column(sa.String(255), nullable=False, index=True)
```
nmalfroy, line 54: Why do you need all of these indexes?  That seems like it's going to take a ton of space and I'm not sure you need them?

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -17,14 +19,20 @@ class CellOperationsDataManager(BaseDataManager):
...
 
     # SQL template file paths
-    SQL_MATCH_METADATA = f"{CELL_ANALYSIS_TEMPLATE_DIR}/get_neighborhood_distance_summary.sql.mako"
+    SQL_MATCH_METADATA = f"{CELL_OPERATIONS_TEMPLATE_DIR}/get_neighborhood_distance_summary.sql.mako"
     SQL_MATCH_METADATA_DEV_DETAILS = (
-        f"{CELL_ANALYSIS_TEMPLATE_DIR}/get_neighborhood_distance_summary_dev_details.sql.mako"
+        f"{CELL_OPERATIONS_TEMPLATE_DIR}/get_neighborhood_distance_summary_dev_details.sql.mako"
     )
-    SQL_GET_CELLS_BY_IDS = f"{CELL_ANALYSIS_TEMPLATE_DIR}/get_cell_metadata_by_ids.sql.mako"
+    SQL_GET_CELLS_BY_IDS = f"{CELL_OPERATIONS_TEMPLATE_DIR}/create_temp_table_for_metadata.sql.mako.mako"
```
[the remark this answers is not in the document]
fedorgrab: Oh this one I will have to remove, it's redundant. But you're right, even before it wasn't supposed to be `.mako.mako`.:)

src/casp/services/db/models/cells.py
fedorgrab: I added indexes to all the columns that I want to have filters by in the admin page. Otherwise querying from admin becomes slow.

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -8,7 +8,7 @@
 from casp.services import settings
 from casp.services.api.clients.matching_client import MatchResult
 from casp.services.api.data_manager import bigquery_response_parsers, bigquery_schemas, cellarium_general
-from casp.services.db import models
+from casp.services.db.models import CASModel, CellInfo
```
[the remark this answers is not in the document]
fedorgrab: Oh I made it because in my older version I was using the models quite often, I will turn it back to just `from casp.services.db import models`

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -25,6 +25,7 @@ class CellOperationsDataManager(BaseDataManager):
         f"{CELL_ANALYSIS_TEMPLATE_DIR}/get_neighborhood_distance_summary_dev_details.sql.mako"
     )
     SQL_GET_CELLS_BY_IDS = f"{CELL_ANALYSIS_TEMPLATE_DIR}/get_cell_metadata_by_ids.sql.mako"
+    redis_cache_key_prefix = "cell_operations"
```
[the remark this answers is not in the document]
fedorgrab: Nope that's redundant, thanks!

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,52 @@ def get_neighborhood_distance_summary_dev_details(
...
-        )
-        sql_query = sql.render(template_path=self.SQL_GET_CELLS_BY_IDS, template_data=template_data)
-
-        query_job = self.block_coo_matrix_db_client.query(query=sql_query)
-
-        return bigquery_response_parsers.parse_get_cells_job(
-            query_job=query_job, cell_metadata_features=metadata_feature_names
-        )
+        # Attempt to get cell metadata from cache
+        cache_keys = [f"cell_metadata:{cell_id}" for cell_id in cell_ids]
```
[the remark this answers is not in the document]
fedorgrab: I believe I can just resolve this comment as this is from the older commit. Sorry for not saying that file change review is preferable over the commit by commit review.:)

src/casp/services/db/migrations/versions/0010_add_cell_related_models.py
```diff
@@ -0,0 +1,91 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0010"
+down_revision = "0009"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
```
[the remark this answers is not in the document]
fedorgrab: Resolving as I think it's solved in the newer commit.:)

src/casp/services/db/migrations/versions/0010_add_cell_related_models.py
```diff
@@ -0,0 +1,91 @@
...
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("cas_feature_index"),
+    )
+    op.create_table(
+        "cells_cellinfo",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("cas_cell_index", sa.Integer(), nullable=False),
+        sa.Column("cas_ingest_id", sa.String(length=255), nullable=False),
+        sa.Column("original_cell_id", sa.String(length=255), nullable=False),
+        sa.Column("obs_metadata_extra", sa.JSON(), nullable=False),
```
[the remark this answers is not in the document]
fedorgrab: I think I found the way to do it, yep, I will try it and will get back here.

src/casp/services/db/models/cells.py
nmalfroy: Hmm...I'm wondering if it's worth it.  Do we really need that view in the admin UI and does it need to be performant?

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -12,12 +13,15 @@
 
 
 @cell_operations_router.post(
-    "/annotate", response_model=t.Union[t.List[schemas.QueryCellDevDetails], t.List[schemas.QueryCell]]
+    "/annotate",
+    response_model=schemas.QueryAnnotationType,
```
nmalfroy, line 17: We talked offline but this should be a separate endpoint since it returns a different type of object

src/casp/services/db/models/cells.py
fedorgrab: Well, I kinda think it is worth it. I myself will use this for querying the database from admin and explore it, or using raw sql. Especially, if we're moving to the "postgres first" idea, I've been making those queries a lot in bigquery. Also I think it can be very helpful to other ppl from Cellarium. 

Also, all these additional indexes together take about 5.2 gb space. The the data itself takes about 52gb. So it's 10% of the data size.

fedorgrab: Are there any performance concerns?

src/casp/data_manager/base_data_manager.py
```diff
@@ -25,3 +25,4 @@ def __init__(self):
         self.project = project
         self.block_coo_matrix_db_client = bigquery.Client(credentials=self.credentials, project=self.project)
         self.system_data_db_session = init_db()
+        self.system_data_db_session.rollback()
```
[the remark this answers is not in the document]
fedorgrab: Oh, right.

src/casp/services/_settings.py
```diff
@@ -45,6 +45,8 @@ class AllEnvSettings(BaseSettings):
     GET_MATCHES_RETRY_BACKOFF_MULTIPLIER: int = 2
     GET_MATCHES_RETRY_BACKOFF_MIN: int = 0
     GET_MATCHES_RETRY_BACKOFF_MAX: int = 30
+    # Consensus Engine
+    GCS_CELL_ONTOLOGY_RESOURCE_FILE: str = "gs://cellarium-file-system/consensus_engine/cell_ontology_resources.json"
```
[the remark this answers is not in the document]
fedorgrab: Can you please explain what do you mean by configuration? I thought settings is already a configuration module.

src/casp/services/api/data_manager/cell_operations.py
fedorgrab: Do you mean we need to factor this out to settings?

src/casp/services/api/data_manager/cell_operations.py
fedorgrab: Hmmm, I think it's a great idea. Let me try it to do it within this PR, I believe it should be pretty simple. I think we already have custom data related error class, so I can just raise it and handle it service. There's a mechanism that handles service errors which can wrap service errors to HTTP 4XX responses to the user with the error message body in detail.

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -54,15 +54,12 @@ async def nearest_neighbor_search(
...
-    user: models.User = Depends(dependencies.authenticate_user),
+    # user: models.User = Depends(dependencies.authenticate_user),
 ):
     """
     Get cells by their ids from a single anndata file with Cellarium CAS. Input file should be validated and sanitized
     according to the model schema.
     """
     return cell_operations_service.get_cells_by_ids_for_user(
         cell_ids=item.cas_cell_ids,
-        model_name=item.model_name,
```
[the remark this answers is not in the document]
fedorgrab: Now, we don't need a model name to get the cells. Previously we did  need it because we had bigquery datasets and we had different datasets for different models. Now we have only one main cell info repository, so we don't need to add model_name to get cells.

src/casp/services/api/data_manager/bigquery_response_parsers.py
fedorgrab: At the moment it still does parses the bigqeury responses. Once we rewrite CellTypeCount based annotation, we will have to get rid of these files.

src/casp/services/db/models/cells.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class CellIngestInfo(db.Base):
+    id = sa.Column(sa.Integer, primary_key=True)
+    cas_ingest_id = sa.Column(sa.String(255), unique=True, nullable=False)
+    dataset_id = sa.Column(sa.String(255), nullable=False)
+    uns_metadata = sa.Column(sa.JSON, nullable=False, default={})
+    ingest_timestamp = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+
+    __tablename__ = "cells_cellingestinfo"
```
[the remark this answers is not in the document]
fedorgrab: Totally makes sense!

src/casp/services/db/models/cells.py
nmalfroy: fair enough!  The performance concerns are for loading the data.  I think with the pipeline that loads these, we'll want to disable the indexes, load, then reenable the indexes (generally good practice for bigger loads)

src/casp/services/db/migrations/versions/0010_add_cell_related_models.py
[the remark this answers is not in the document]
fedorgrab: Yeah, seems like it was a very easy fix.

src/casp/services/api/data_manager/bigquery_response_parsers.py
nmalfroy: Ah ok gotcha

src/casp/services/api/data_manager/cell_operations.py
nmalfroy: Yeah exactly.

src/casp/services/api/routers/cell_operations_router.py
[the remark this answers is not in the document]
nmalfroy: Got it.  And we don't need user information?

src/casp/services/_settings.py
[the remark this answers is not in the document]
nmalfroy: oh sorry, I confused myself as I was typing.  I was thinking it should just be:
`consensus_engine/cell_ontology_resources.json` and that should be appended to the PROJECT_BUCKET_NAME value when reading.  It seems odd to have some variables use a full gs path and others be descendants of a bucket.  Does that make sense?

src/casp/services/api/routers/cell_operations_router.py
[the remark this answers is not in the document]
fedorgrab: Nope, because previously, we used user only to validate that the model is available for the user (e.g. when model was only for admin users, we checked that it was the admin user calling it), but now there is no model to validate for the user, so no user needed as well.

src/casp/services/_settings.py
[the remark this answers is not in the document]
fedorgrab: Ohhhh right! Yeah, that totally makes sense.

nmalfroy:
Meta question: how does the data get ingested into postgres?  We should document that in the readme

src/casp/services/api/routers/cell_operations_router.py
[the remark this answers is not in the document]
nmalfroy: oh awesome! Cool

src/casp/services/db/models/cells.py
nmalfroy: Yep, that's exactly right

src/casp/services/db/models/users.py
```diff
@@ -12,7 +12,7 @@ class User(db.Base):
     cells_processed = sa.Column(sa.Integer, default=0, nullable=False)
     is_admin = sa.Column(sa.Boolean(), default=True, nullable=False)
 
-    __tablename__ = "user"
+    __tablename__ = "users_user"
```
nmalfroy, line 15: Sorry, if you're renaming, should this just be `users`?

fedorgrab:
> Meta question: how does the data get ingested into postgres? We should document that in the readme

About that, we have cell data lifecycle scripts, one of which is 'ingest'. I believe there should be significant refactoring there (they've been used for bq ingest). Also, this seems like it should be a separate PR to me.

nmalfroy:
> > Meta question: how does the data get ingested into postgres? We should document that in the readme
> 
> About that, we have cell data lifecycle scripts, one of which is 'ingest'. I believe there should be significant refactoring there (they've been used for bq ingest). Also, this seems like it should be a separate PR to me.

Yep, totally agreed.  Just wanted to make sure it's on the radar

src/casp/services/db/models/users.py
fedorgrab: Let me give you some context and motivation behind it. The naming convention I've adopted was inspired by the database table naming policy from the Django framework. The format I use is 'db_model_module_ormModelName'. This approach allows us to easily structure tables by names and application modules. Our current application module is 'users' (where all user-related models go), and the model name is 'User'; thus, the table name is 'users_user'. I thought it would be useful to incorporate table prefixes. This way, when you add your new model to auth-related stuff (e.g., Token or ServiceAccount), the table name appears as 'users_token' or 'users_serviceaccount', respectively. Also it helps to organize tables when listing them from data grips (screenshot attached). Please let me know what you think about it.
image

fedorgrab:
> > > Meta question: how does the data get ingested into postgres? We should document that in the readme
> > 
> > 
> > About that, we have cell data lifecycle scripts, one of which is 'ingest'. I believe there should be significant refactoring there (they've been used for bq ingest). Also, this seems like it should be a separate PR to me.
> 
> Yep, totally agreed. Just wanted to make sure it's on the radar

Oh totally there's [an issue](https://github.com/cellarium-ai/cellarium-cloud/issues/94) for it.

src/casp/services/db/models/users.py
nmalfroy: ah ok.  I got you.  I think that's fine.  And it's definitely better than `user` which was causing syntax errors in SQL if left unquoted

nmalfroy (approved):

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
+        self, cell_ids: t.List[int], feature_names: t.List[str]
+    ) -> t.List[schemas.CellariumCellMetadata]:
         """
-        Get cells by ids.
+        Get cells by ids, maintaining the order of cell_ids. Preserves the order of cell_ids in the query results.
 
-        :param cell_ids: Cas cell indexes from Big Query
-        :param metadata_feature_names: Metadata features to return from Big Query `cas_cell_info` table
-        :param model_name: Name of the model to query. Used to get the dataset name where to get the cells from
+        :param cell_ids: Cas cell indexes to retrieve metadata for.
```
mbabadi, line 121: Cas -> CAS

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
 
-    def get_cell_metadata_by_ids(
-        self, cell_ids: t.List[int], metadata_feature_names: t.List[str], model_name: str
-    ) -> t.List[t.Dict[str, t.Any]]:
+    def _get_cell_metadata_by_ids(
+        self, cell_ids: t.List[int], feature_names: t.List[str]
+    ) -> t.List[schemas.CellariumCellMetadata]:
         """
-        Get cells by ids.
+        Get cells by ids, maintaining the order of cell_ids. Preserves the order of cell_ids in the query results.
```
mbabadi, line 119: "Get cells by ids ..." -> "Get cells by CAS IDs, maintaining the order of ``cell_ids``. Preserves the order of ``cell_ids``..."

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
+        index_to_metadata = {row["cas_cell_index"]: row for row in query_result}
+        ordered_metadata = (index_to_metadata[cell_id] for cell_id in cell_ids if cell_id in index_to_metadata)
+
+        return [schemas.CellariumCellMetadata(**cell_metadata_row) for cell_metadata_row in ordered_metadata]
+
+    def get_cell_metadata_by_ids(
+        self, cell_ids: t.List[int], metadata_feature_names: t.List[str]
+    ) -> t.List[schemas.CellariumCellMetadata]:
+        """
+        Get cells by ids, maintaining the order of cell_ids. If the number of querying cell IDs exceeds the maximum
```
mbabadi, line 163: Get cell metadata by CAS IDs, maintaining the order of ``cell_ids``.

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
+        allowed, split the query into multiple queries. Database gets overwhelmed when querying a large number of cell.
+
+        :param cell_ids: Cas cell indexes to retrieve metadata for.
```
mbabadi, line 166: "Cas cell indexes" -> "CAS cell IDs".

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
+        :param metadata_feature_names: List of feature names to retrieve.
+
+        :return: List of dictionaries representing the query results, ordered according to cell_ids.
```
mbabadi, line 169: cell_ids -> ``cell_ids``

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
+        """
+        if len(cell_ids) >= settings.MAX_CELL_IDS_PER_QUERY:
+            neighbors_metadata = []
+            for i in range(0, len(cell_ids), settings.MAX_CELL_IDS_PER_QUERY):
+                neighbors_metadata += self._get_cell_metadata_by_ids(
+                    cell_ids=list(cell_ids)[i : i + settings.MAX_CELL_IDS_PER_QUERY],
```
mbabadi, line 175: why do `list(cell_ids)[slice]` and not just `cell_ids[slice]`?

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
+                    feature_names=metadata_feature_names,
+                )
+        else:
+            neighbors_metadata = self._get_cell_metadata_by_ids(
+                cell_ids=list(cell_ids), feature_names=metadata_feature_names
```
mbabadi, line 180: Why `list(cell_ids)`?

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -11,13 +12,11 @@
 cellarium_general_service = services.CellariumGeneralService()
 
 
-@cell_operations_router.post(
-    "/annotate", response_model=t.Union[t.List[schemas.QueryCellDevDetails], t.List[schemas.QueryCell]]
-)
+@cell_operations_router.post("/annotate", response_model=schemas.QueryAnnotationCellTypeCountType)
```
mbabadi, line 15: I think it is cleaner to have a single `/annotate` URL with the strategy field, rather than `/annotate` and `/annotate-ontology-aware`.

nmalfroy: Normally I would agree but the problem is that the parameters changes the response shape. It's generally better practice to have that in a separate api call in that case since dealing with polymorphism on the client is a pretty bad experience for users

src/casp/services/api/data_manager/cell_operations.py
fedorgrab: That's a good point, seems like it's a mistake.

src/casp/services/api/schemas/cell_operations.py
```diff
@@ -22,18 +22,46 @@ class DevDetailObject(BaseModel):
     mean_distance: float = Field(example=1791.372802734375)
 
 
-class MatchInfoDevDetails(MatchInfo):
+class AnnotationInfoCellTypeCountDevDetail(AnnotationInfoCellTypeCount):
```
mbabadi, line 25: `NeighborhoodSummaryCellTypeStatiticsExtended`.  I was never a fan of `DevDetail`.  While you're at it, please rename those too.

src/casp/services/api/schemas/cell_operations.py
```diff
@@ -22,18 +22,46 @@ class DevDetailObject(BaseModel):
...
     dataset_ids_with_counts: t.List[DevDetailObject]
 
 
-class QueryCell(BaseModel):
+class QueryCellAnnotationAbstract(BaseModel):
```
mbabadi, line 29: Call it `QueryCellNeighborhoodAbstract`.

Annotation is a task we're trying to perform, but something we're getting from the query. Correct naming matters :)

src/casp/services/api/data_manager/bigquery_response_parsers.py
```diff
@@ -33,30 +35,30 @@ def _parse_match_object_dev_details(row: bigquery.Row) -> t.Dict[str, t.Any]:
...
+        p25_distance=row["p25_distance"],
+        median_distance=row["median_distance"],
+        p75_distance=row["p75_distance"],
+        max_distance=row["max_distance"],
+        dataset_ids_with_counts=dataset_ids_with_counts,
+    )
 
 
 def parse_match_query_job(
     query_job: bigquery.QueryJob, include_dev_details: bool = False
```
mbabadi, line 60: `include_dev_details` -> `generate_extended_output`

src/casp/services/api/data_manager/cell_operations.py
```diff
@@ -111,31 +112,72 @@ def get_neighborhood_distance_summary_dev_details(
...
+        :param cell_ids: CAS cell IDs to retrieve metadata for.
+        :param metadata_feature_names: List of feature names to retrieve.
+
+        :return: List of dictionaries representing the query results, ordered according to `cell_ids`.
+        """
+        if len(cell_ids) >= settings.MAX_CELL_IDS_PER_QUERY:
+            neighbors_metadata = []
+            for i in range(0, len(cell_ids), settings.MAX_CELL_IDS_PER_QUERY):
+                neighbors_metadata += self._get_cell_metadata_by_ids(
+                    cell_ids=cell_ids[i : i + settings.MAX_CELL_IDS_PER_QUERY],
```
mbabadi, line 177: Does it make sense to replace `i + settings.MAX_CELL_IDS_PER_QUERY` with `min(len(cell_ids), i + settings.MAX_CELL_IDS_PER_QUERY)`? If `cell_ids` is either a list or ndarray, we get the right behavior. If it is some other iterable that may behave differently if the upper limit of the slice is above the lenghth, we will be in trouble.

src/casp/services/api/routers/cell_operations_router.py
mbabadi: OK, so let's break backwards compatibility and call this `/annotate-cell-type-count-strategy`. It doesn't make sense a "default" `/annotate`, especially one that we are walking away from soon. Perhaps we could revisit the issue of having a polymorphic `/annotate`. We are likely going to study other types of query summarization strategies soon (e.g. disease, publication, etc). It is fine to have separate API calls for different strategies, but it will quickly go out of control if we have many more strategies. I am not too worried about the client-side UX. The users always get a list of dicts, and they know what kind of dict they are getting based on what strategy they have chosen...

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -11,13 +12,11 @@
...
 async def annotate(
     file: UploadFile = File(),
     model_name: str = Form(),
-    include_dev_metadata: bool = Form(),
+    include_dev_metadata: t.Optional[bool] = Form(default=None),
```
mbabadi, line 19: `annotate` -> `annotate_cell_type_count_strategy` (no implied default)
`include_dev_metadata` -> `include_extended_metadata` (here and elsewhere)

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -31,8 +30,44 @@ async def annotate(
...
+    :return: JSON response with annotations.
+    """
+
     return await cell_operations_service.annotate_adata_file(
-        user=request_user, file=file.file, model_name=model_name, include_dev_metadata=include_dev_metadata
+        user=request_user,
+        file=file.file,
+        model_name=model_name,
+        consensus_strategy=ConsensusStrategyType.ONTOLOGY_AWARE,
+        normalize=normalize,
```
mbabadi, line 70: I suggest dropping `normalize` altogether for clarity. Also, the docstring is wrong (`normalize the data before annotation`), which raised my eyebrows for a second :-)

We want the scores, by definition, to range in [0, 1]. It might be a good idea to include `total_weight` as a part of the output in the dictionary:
```
{
    ontology_term_scores = {
        CL_... = ...,
        CL_... = ...,
        CL_... = ...,
        CL_... = ...,
    },
    misc = {
        total_weight = ...,
    }
}
```

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -31,8 +30,44 @@ async def annotate(
...
+    normalize: t.Optional[bool] = Form(default=None),
+    request_user: models.User = Depends(dependencies.authenticate_user),
+):
+    """
+    Annotate a single anndata file with Cellarium CAS. Input file should be validated and sanitized according to the
+    model schema.
+
+    :param file: Byte object of :class:`anndata.AnnData` file to annotate.
+    :param model_name: Model name to use for annotation. See `/list-models` endpoint for available models.
+    :param normalize: Boolean flag indicating whether to normalize the data before annotation.
```
mbabadi, line 59: (drop)

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -31,8 +30,44 @@ async def annotate(
...
+
+
+@cell_operations_router.post(
+    "/annotate-ontology-aware-strategy",
+    response_model=schemas.QueryAnnotationOntologyAwareType,
+)
+async def annotate_ontology_aware_strategy(
+    file: UploadFile = File(),
+    model_name: str = Form(),
+    normalize: t.Optional[bool] = Form(default=None),
```
mbabadi, line 50: (drop)

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -170,67 +170,71 @@ def __split_embeddings_into_chunks(embeddings: np.array, chunk_size: int) -> t.L
...
-        return self.cell_operations_dm.get_neighborhood_distance_summary(
-            cas_model=cas_model, match_temp_table_fqn=temp_table_fqn
-        )
-
-    async def annotate_adata_file(
-        self, user: models.User, file: t.BinaryIO, model_name: str, include_dev_metadata: bool
-    ) -> t.List[t.Dict[str, t.Any]]:
+        consensus_strategy: consensus_engine.ConsensusStrategyType,
+        include_dev_metadata: t.Optional[bool] = None,
+        normalize: t.Optional[bool] = None,
```
mbabadi, line 180: (drop)

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -170,67 +170,71 @@ def __split_embeddings_into_chunks(embeddings: np.array, chunk_size: int) -> t.L
...
-            model_name=model_name,
-            include_dev_metadata=include_dev_metadata,
-        )
+        cas_model = self.cellarium_general_dm.get_model_by_name(model_name=model_name)
+
+        match consensus_strategy:
+            case consensus_engine.ConsensusStrategyType.ONTOLOGY_AWARE:
+                strategy = consensus_engine.OntologyAwareConsensusStrategy(
+                    cell_operations_dm=self.cell_operations_dm,
+                    normalize=normalize,
```
mbabadi, line 218: (drop)

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+    assigned to each neighbor cell type based on their distance and cell type ontology, and the weights are propagated
+    to their ancestors in the cell type ontology graph.
+    Algorithm:
+
+    1. Get Metadata for each unique neighbor cell.
+    2. Iterate over each query cell.
+       2.1. Get distances for each neighbor cell.
+       2.2. Calculate weights for each neighbor cell.
+       2.3. Update weights for each neighbor cell and their ancestors in the cell type ontology.
+       2.4. Normalize the weights if required.
```
mbabadi, line 120: (drop)

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+            # Add weight to the neighbor's ontology ID
+            scores_dict[neighbor_cell_type_ontology_id] += weight
+
+            # Propagate the weight to all ancestors (consistent subgraph of the cell type ontology)
+            ancestor_ontology_ids = self.cell_ontology_resource.ancestors_dictionary[neighbor_cell_type_ontology_id]
+
+            for ancestor in ancestor_ontology_ids:
+                scores_dict[ancestor] += weight
+
+        if self.normalize:
```
mbabadi, line 185: (drop)

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+        if self.prune_threshold > 0:
+            scores_dict = {k: v for k, v in scores_dict.items() if v >= self.prune_threshold}
+
+        return [
+            schemas.NeighborhoodSummaryOntologyAware(
+                score=score,
+                cell_type_ontology_term_id=cell_type_ontology_term_id,
+                cell_type=self.cell_ontology_resource.ontology_term_id_to_name_dict[cell_type_ontology_term_id],
+            )
+            for cell_type_ontology_term_id, score in scores_dict.items()
```
mbabadi, line 197: also return `total_weight` and `total_usable_neighbors`.

src/casp/services/api/schemas/cell_operations.py
```diff
@@ -3,7 +3,7 @@
 from pydantic import BaseModel, Field
 
 
-class MatchInfo(BaseModel):
+class NeighborhoodSummaryCellTypeStatistics(BaseModel):
```
mbabadi, line 6: `NeighborhoodSummaryCellTypeStatistics` -> ``NeighborhoodCellTypeSummaryStatistics`. "Summary Statistics" is a mathematical term, e.g. median, p25, etc are all "summary statistics".

src/casp/services/api/schemas/cell_operations.py
```diff
@@ -22,18 +22,46 @@ class DevDetailObject(BaseModel):
     mean_distance: float = Field(example=1791.372802734375)
 
 
-class MatchInfoDevDetails(MatchInfo):
-    dataset_ids_with_counts: t.List[DevDetailObject]
+class NeighborhoodSummaryCellTypeStatisticsExtended(NeighborhoodSummaryCellTypeStatistics):
```
mbabadi, line 25: summary statistics

src/casp/services/api/schemas/cell_operations.py
```diff
@@ -22,18 +22,46 @@ class DevDetailObject(BaseModel):
...
 
-class QueryCell(BaseModel):
+class QueryCellNeighborhoodAbstract(BaseModel):
     query_cell_id: str = Field(example="ATTACTTATTTAGTT-12311")
-    matches: t.List[MatchInfo]
+    matches: t.List
 
 
-class QueryCellDevDetails(BaseModel):
+class QueryCellNeighborhoodCellTypeCount(QueryCellNeighborhoodAbstract):
```
mbabadi, line 34: QueryCellNeighborhoodCellTypeSummaryStatistics?

src/casp/services/api/schemas/cell_operations.py
```diff
@@ -22,18 +22,46 @@ class DevDetailObject(BaseModel):
...
+    matches: t.List[NeighborhoodSummaryCellTypeStatistics]
+
+
+class QueryCellNeighborhoodCellTypeCountExtended(QueryCellNeighborhoodAbstract):
```
mbabadi, line 38: QueryCellNeighborhoodCellTypeSummaryStatisticsExtended?

src/casp/services/api/routers/cell_operations_router.py
nmalfroy: Hmm, not sure I agree with that.  By that description, why bother having any structure at all?  These are fundamentally different objects being returned so it seems reasonable to my to have them be different API calls.  I think returning arbitrary json with no help to the user is not a pattern I feel particularly good about promoting.

If we're going to have an arbitrary amount of these, we should think about what type of response could encapsulate all of these but honestly, having an API for each doesn't seem all that bad to me.

As for breaking backwards compatibility, I don't know that we need to do that.  We could have annotate be deprecated and then migrate 10x to the new version. Just trying not to pull the rug out from under the,/

nmalfroy: Also worth talking about in person...I tend to come off as a jerk when I have these conversations written down and definitely don't mean to!

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+    Attributes:
+        ancestors_dictionary: Maps cell ontology term IDs to their ancestor IDs.
+        ontology_term_id_to_name_dict: Maps cell ontology term IDs to cell type names.
+
+    Raises:
+        ValueError: If either ``ancestors_dictionary`` or ``cell_ontology_term_id_to_cell_type`` is missing from the
+            provided dictionary.
+    """
+
+    def __init__(self, cell_ontology_resource_dict: t.Optional[t.Dict[str, t.Any]] = None):
```
mbabadi, line 92: include cell ontology version that is used, and a permalink to where the ontology obo file is located.

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+class OntologyAwareConsensusStrategy(ConsensusStrategyProtocol):
+    """
+    Handle ontology-aware consensus strategy, summarizing query neighbor context by cell type ontology. Weights are
+    assigned to each neighbor cell type based on their distance and cell type ontology, and the weights are propagated
+    to their ancestors in the cell type ontology graph.
+    Algorithm:
+
+    1. Get Metadata for each unique neighbor cell.
```
mbabadi, line 115: Metadata -> metadata

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+    :param prune_threshold: Threshold for pruning weights below a certain value. If 0, no pruning is performed.
+    :param cell_ontology_resource: Cell ontology resource object.
+    :param cell_operations_dm: Cell operations data manager object.
+    """
+
+    REQUIRED_CELL_INFO_FEATURE_NAMES = ["cas_cell_index", "cell_type", "cell_type_ontology_term_id"]
+
+    def __init__(
+        self,
+        normalize: bool = True,
```
mbabadi, line 132: drop

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+        prune_threshold: float = 0.1,
```
mbabadi, line 133: set the default prune threshold to 0.

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+        Utilize the ontology-aware method to assign weights to neighbor cells based on their distance and cell type
+        ontology, to inform context summarization.
+
+        :param neighbors: Neighbors of the query cell as determined by the matching engine.
+        :param neighbors_metadata_dict: Metadata for each neighbor cell.
+        :return: A list of `schemas.AnnotationInfoOntologyAware` instances, each representing the weighted cell type
+             ontology term for a neighbor cell.
+        """
+
+        neighbor_distances = np.asarray([neighbor.distance for neighbor in neighbors])
```
mbabadi, line 157: We need to be careful with distances and similarity scores. It looks like Vertex AI is returning cosine similarity scores. The distance, in that case, is typically defined as `1 - cossim(x, y)`. If we are storing _distance_ as an attribute of a match, we need to make sure that it is indeed a distance. My workaround has been to change the sign of `gamma`, which is a terrible idea. The better thing to do is to fix this upstream. Can you check what is going on? Is Vertex AI actually returning similarity scores and calling it distance, or have we called it distance by mistake? Does Vertex AI always return similarity scores regardless of what distance metric is used (likely), or returns distance for some metrics and similarity scores for other metrics (unlikely)?

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+
+        neighbor_metadata = [neighbors_metadata_dict[neighbor.cas_cell_index] for neighbor in neighbors]
+
+        # Get weights for each neighbor
+        gamma = 1.0 / np.median(neighbor_distances)
```
mbabadi, line 162: Expose the prefactor of gamma (currently 1.0), and call it `distance_exponential_weighting_prefactor`. Expose it all the way to the API level. When you figured out the issue with distance/similarity and ensured that indeed we're working with distances, put a minus sign before the definition of gamma:
```
gamma = -distance_exponential_weighting_prefactor / np.median(neighbor_distances)
```

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+        total_weight = 0
+        total_usable_neighbors = 0
+        scores_dict = {k: 0 for k in self.cell_ontology_resource.ancestors_dictionary.keys()}
+
+        for neighbor_metadata, weight in zip(neighbor_metadata, weights):
+            # Cell Ontology IDs have the formats: CL:0000000 (in our database) and CL_0000000 (in the ontology graph)
+            neighbor_cell_type_ontology_id = neighbor_metadata.cell_type_ontology_term_id.replace(":", "_")
+
+            total_weight += weight
+            total_usable_neighbors += 1
```
mbabadi, line 174: rename to `total_neighbors`.

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+            for ancestor in ancestor_ontology_ids:
+                scores_dict[ancestor] += weight
+
+        if self.normalize:
+            scores_dict = {k: v / total_weight for k, v in scores_dict.items()}
+
+        if self.prune_threshold > 0:
+            scores_dict = {k: v for k, v in scores_dict.items() if v >= self.prune_threshold}
+
+        return [
```
mbabadi, line 191: return `total_weight` and `total_neighbors` as well (see my other comment about this in the router).

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+            schemas.NeighborhoodSummaryOntologyAware(
+                score=score,
+                cell_type_ontology_term_id=cell_type_ontology_term_id,
+                cell_type=self.cell_ontology_resource.ontology_term_id_to_name_dict[cell_type_ontology_term_id],
+            )
+            for cell_type_ontology_term_id, score in scores_dict.items()
+        ]
+
+    def summarize(self, query_ids: t.List[str], knn_query: MatchResult) -> schemas.QueryAnnotationOntologyAwareType:
```
mbabadi, line 200: expose `prune_threshold` and `distance_exponential_weighting_prefactor` here.

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+            neighbor_cell_type_ontology_id = neighbor_metadata.cell_type_ontology_term_id.replace(":", "_")
+
+            total_weight += weight
+            total_usable_neighbors += 1
+
+            # Add weight to the neighbor's ontology ID
+            scores_dict[neighbor_cell_type_ontology_id] += weight
+
+            # Propagate the weight to all ancestors (consistent subgraph of the cell type ontology)
+            ancestor_ontology_ids = self.cell_ontology_resource.ancestors_dictionary[neighbor_cell_type_ontology_id]
```
mbabadi, line 180: intersting that the dictionary is working faster than array slicing...

src/casp/services/api/services/consensus_engine/strategies.py
mbabadi: use kwargs if you want to keep the base class simple. in that case, put the default values here and not in the inner function.

src/casp/services/api/services/cell_operations_service.py
mbabadi: instead, add **kwargs to allow passing strategy-specific parameters (like pruning thershold, prefactor).

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -170,67 +170,71 @@ def __split_embeddings_into_chunks(embeddings: np.array, chunk_size: int) -> t.L
...
-        :param include_dev_metadata: Boolean flag indicating whether to include dev metadata in the response.
+        :param consensus_strategy: Consensus strategy to use for annotation from `ConsensusStrategyType` enum.
+        :param normalize: Boolean flag indicating whether to normalize the embeddings before annotation. Used only in
+            `ontology_aware` method.
+        :param include_dev_metadata: Boolean flag indicating whether to include dev metadata in the response. Used only
+            in `cell_type_count` method.
 
         :return: JSON response with annotations.
         """
+        if normalize is None and consensus_strategy == consensus_engine.ConsensusStrategyType.ONTOLOGY_AWARE:
```
mbabadi, line 197: drop

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -170,67 +170,71 @@ def __split_embeddings_into_chunks(embeddings: np.array, chunk_size: int) -> t.L
...
+            case consensus_engine.ConsensusStrategyType.ONTOLOGY_AWARE:
+                strategy = consensus_engine.OntologyAwareConsensusStrategy(
+                    cell_operations_dm=self.cell_operations_dm,
+                    normalize=normalize,
+                )
+            case consensus_engine.ConsensusStrategyType.CELL_TYPE_STATISTICS:
+                strategy = consensus_engine.CellTypeStatisticsConsensusStrategy(
+                    cell_operations_dm=self.cell_operations_dm,
+                    cas_model=cas_model,
+                    include_dev_metadata=include_dev_metadata,
```
mbabadi, line 224: this could be a kwarg for CELL_TYPE_STATISTICS strategy w/ default value set to False.

src/casp/services/db/models/cells.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class CellIngestInfo(db.Base):
+    id = sa.Column(sa.Integer, primary_key=True)
+    cas_ingest_id = sa.Column(sa.String(255), unique=True, nullable=False)
+    dataset_id = sa.Column(sa.String(255), nullable=False)
+    uns_metadata = sa.Column(JSONB, nullable=False, default={})
+    ingest_timestamp = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+
+    __tablename__ = "cells_ingestinfo"
```
mbabadi, line 17: In BQ, we are calling that `cas_ingest_info`. Why are we calling this `cells_ingestinfo`?

src/casp/services/db/models/cells.py
```diff
@@ -0,0 +1,78 @@
...
+    feature_name = sa.Column(sa.String(255), nullable=False)
+    feature_biotype = sa.Column(sa.String(255), nullable=True)
+    feature_is_filtered = sa.Column(sa.Boolean(), default=False, nullable=True)
+    feature_reference = sa.Column(sa.String(255), nullable=True)
+    var_metadata_extra = sa.Column(JSONB, nullable=False, default={})
+    cas_ingest_id = sa.Column(
+        sa.String(255), sa.ForeignKey(f"{CellIngestInfo.__tablename__}.cas_ingest_id"), nullable=False
+    )
+
+    __tablename__ = "cells_featureinfo"
```
mbabadi, line 36: same comment as above

src/casp/services/db/models/cells.py
```diff
@@ -0,0 +1,78 @@
...
+    cell_type_ontology_term_id = sa.Column(sa.String(255), nullable=True, index=True)
+    assay_ontology_term_id = sa.Column(sa.String(255), nullable=True)
+    development_stage_ontology_term_id = sa.Column(sa.String(255), nullable=True)
+    tissue_ontology_term_id = sa.Column(sa.String(255), nullable=True)
+    disease_ontology_term_id = sa.Column(sa.String(255), nullable=True)
+    organism_ontology_term_id = sa.Column(sa.String(255), nullable=True)
+    self_reported_ethnicity_ontology_term_id = sa.Column(sa.String(255), nullable=True)
+    sex_ontology_term_id = sa.Column(sa.String(255), nullable=True)
+
+    __tablename__ = "cells_cellinfo"
```
mbabadi, line 75: same comment as above

src/casp/services/db/models/ml_management.py
```diff
@@ -53,4 +38,4 @@ class CASMatchingEngineIndex(db.Base):
     def __str__(self):
         return self.index_name
 
-    __tablename__ = "cas_matching_engine_index"
+    __tablename__ = "ml_management_matchingengineindex"
```
mbabadi, line 41: why no `_`s in `machineengineindex`?

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+        ontology_term_id_to_name_dict: Maps cell ontology term IDs to cell type names.
+
+    Raises:
+        ValueError: If either ``ancestors_dictionary`` or ``cell_ontology_term_id_to_cell_type`` is missing from the
+            provided dictionary.
+    """
+
+    def __init__(self, cell_ontology_resource_dict: t.Optional[t.Dict[str, t.Any]] = None):
+        if cell_ontology_resource_dict is None:
+            with open(settings.GCS_CELL_ONTOLOGY_RESOURCE_FILE, "rb") as f:
```
mbabadi, line 94: Cell ontology is actually a rapidly evolving resource:
https://github.com/obophenotype/cell-ontology

Here are the releases:
https://github.com/obophenotype/cell-ontology/releases

CELLxGENE is expected to keep migrating to newer versions, and they are actively involved in extending this ontology. While it is unlikely that they update existing cell type annotations for datasets they have already ingested (though it might happen, and we should ask them how to some updates from them), newly ingested datasets will use newer versions of their schema which would use newer cell ontology releases, which would include both new terms and new relationships. Each specific version of the CELLxGENE schema is pegged to a specific release date of the cell ontology. Cell ontology releases are tagged by their release date.

How can we use combine queries across different referrence datasets that use different cell ontology version? It is likely not a very fundamental problem, since most (if not all) new ontology releases are extensions of previous releases. Probably some terms will be deprecated and eventually removed. A proper solution here is to have a cell ontology migration tool (e.g. to replace deprecated terms with preferred terms), so that we migrate all of the data in our DataStore to the latest cell ontology release. Another solution would be to perform migration on-the-fly here (if we know neighbor X comes from dataset Y with cell ontology Z, we can migrate the ontology metadata of that neighbor to the newest cell ontology release on-the-fly). Let's make an issue for this for further discussions and settle for a simpler solution here (see below).

Let us do the following:

- Include scripts in the repo (or better yet, a protected API) to generate the parsed cell ontology JSON from a given versioned release and store it here:
```
gs://{PROJECT_BUCKET_NAME}/consensus_engine/parsed_cell_ontology_release_vYYYY-MM-DD.json
```
All versioned releases can be downloaded from:
```
https://github.com/obophenotype/cell-ontology/releases/download/vYYYY-MM-DD/cl.owl
```

- We expose which version of the cell ontology is to be used in `CellOntologyResource`. If the (parsed) version exists (`gs://{PROJECT_BUCKET_NAME}/consensus_engine/parsed_cell_ontology_release_vYYYY-MM-DD.json`) we proceed. Otherwise, we return an exception. In the future, we can register parsed cell ontologies as a separate table + API call to let the users know which versions of the ontology we have in store (make an issue).

- When we run the consensus-aware scoring, it is possible that we may not have some of the ontology terms in the loaded resource (e.g. if we are using an older resource and but the neighbors are coming from a newer dataset using a newer resource). For now, we ignore those neighbors and keep track of the number of ignored neighbors (see below).

src/casp/services/api/services/consensus_engine/strategies.py
```diff
@@ -0,0 +1,230 @@
...
+
+        for neighbor_metadata, weight in zip(neighbor_metadata, weights):
+            # Cell Ontology IDs have the formats: CL:0000000 (in our database) and CL_0000000 (in the ontology graph)
+            neighbor_cell_type_ontology_id = neighbor_metadata.cell_type_ontology_term_id.replace(":", "_")
+
+            total_weight += weight
+            total_usable_neighbors += 1
+
+            # Add weight to the neighbor's ontology ID
+            scores_dict[neighbor_cell_type_ontology_id] += weight
```
mbabadi, line 177: Check if `neighbor_cell_type_ontology_id` is in the ontology resource. If not, ignore the neighbor and keep track of `total_neighbors_with_unrecognized_cell_ontology_term_id`.

mbabadi (changes requested):
Nice work!  A couple of comments to address (mostly easy).

src/casp/services/api/services/cell_operations_service.py
fedorgrab: I'd like to avoid using `**kwargs` as it makes method arguments less specific. I will add threshold. What's a prefactor? What are other strategy parameters we can have?

src/casp/services/api/services/consensus_engine/strategies.py
fedorgrab: MatchNeighbor object from google's sdk has a variable distance, however in some tutorial notebooks they mention that distance is a similarity. I am going to send them an email and see what they respond. I will leave this discussion unresolved before we make sure it is either distance or a similarity.

src/casp/services/api/services/consensus_engine/strategies.py
fedorgrab: I think it's because getting an element from a dictionary is O(1) while slicing an array is O(n). But this was surprising for me as well, because it was a sparse array, which means that n is not number of elements, but number of non zero elements. But, anyway, I guess even with the sparse array it still affected the computation as it happens in the loop while iterating over a large number of neighbors.

src/casp/services/db/models/ml_management.py
fedorgrab: Please see the response to the previous conversation about that.

src/casp/services/api/services/consensus_engine/strategies.py
fedorgrab: Let's create an issue and work on it through a separate PR. I will include more documentation on cell ontology to keep track of how the resource was created within this PR.

fedorgrab: It just seems like we need to talk more about the versioning of cell ontology and how to better keep track of the records, how this would affect our data and models as well.

nmalfroy (approved):
changes look good!

mbabadi (approved):
Great job! All looks good to me.

PR #150, opened 2024-04-30
No longer include secrets in docker images and improve DevX

nmalfroy:
As part of the work in doing this, I also made some other changes that will benefit security and developer experience.

- Deprecate the `GOOGLE_ACCOUNT_CREDENTIALS` setting.  We should use Google native auth.
   - On cloud run instances, this means that the service account associated with the cloud run instance will be what is used to talk to Google apis (e.g. read models from buckets)
   - On non-GCP instances (e.g. AWS, local workstations) this means setting the `GOOGLE_APPLICATION_CREDENTIALS` environment to the path to a key file (e.g. the JSON that was stored in `GOOGLE_ACCOUNT_CREDENTIALS`)
   - **Note: `GOOGLE_ACCOUNT_CREDENTIALS` is deprecated but still works.  A warning message will be logged when that method is used**
- Created a brand new Github workflow to build the docker images explicitly.  It can be run via the github CLI using:
```
gh workflow run docker-workflow.yml --repo cellarium-ai/cellarium-cloud \
            --ref <branch, git commit, tag, etc.> \
            -f image-types=<standard|cuda|both>
            -f image-tag=<docker tag>
            -f add-latest-tag=<true|false>
```
or via the github UI:
image
    - The github workflow allows you to control how to build the image and emits the published tags on the summary of the run:
image

As a result of this change, configuration of the services will now be mounted to `/app/casp/services/settings/.env` location. 
- When running via docker, this is done by passing in a volume.  E.g. locally it might look like:
```docker run -p 8000:8000 --name local-cas-api -v <path to repo>/src/settings/.env:/app/settings/.env cas-pytorch:test python casp/services/api/main.py```

**Note the image name: `cas-pytorch` as opposed to `cas-pytorch-dev` or `cas-pytorch-prod`. Because we will no longer include config, we no longer need to have different version per environment**

- In Cloud Run deployments, this will be via a secret (e.g.
image

**!!!WARNING!!!**
This PR moves the .env file from `src/casp/services/.env` to `src/casp/services/settings/.env`.  This is because Cloud Run, when mounting secrets into a file, overwrites the child-most directory to mount the secret.  So before the change, the entire contents of `src/casp/services` would be overwritten on launch

Once this is merged, and we've created secrets for dev and prod, we can cut over to this approach and remove the legacy docker builds.  After that we should rotate secrets.

### CODE REVIEW

**Review by KevinCLydon — COMMENTED** [review]
Got a couple very quick questions/suggestions.  Otherwise, this looks great 👍

File: src/casp/services/deploy/Dockerfile.pytorch [hunk 1]
```diff
@@ -20,7 +20,7 @@ RUN pip install -r requirements.txt
 COPY src .
 COPY tests ./tests
 # Looking for b64 encoded secret environment variables
-RUN --mount=type=secret,id=ENV_SECRETS cat /run/secrets/ENV_SECRETS | base64 -d >> ./casp/services/.env
+RUN --mount=type=secret,id=ENV_SECRETS cat /run/secrets/ENV_SECRETS | base64 -d >> ./settings/.env
```
fedorgrab, line 23: Nice backwards compatibility, thanks!

src/settings/sample_env
fedorgrab, line 1: That's very nice to have a sample! MIght be a good idea to rename it to `.env.example` or `.env.sample`.

src/casp/services/_settings.py
```diff
@@ -5,10 +5,15 @@
...
+# Directory that contains the services package
 SERVICES_DIR = os.path.dirname(os.path.abspath(__file__))
+# Directory that contains the CAS package's contents
 CAS_DIR = os.path.dirname(SERVICES_DIR)
+# Directory that contains the CAS package and its settings
+ROOT_DIR = os.path.dirname(CAS_DIR)
 
-dotenv.load_dotenv(dotenv_path=f"{SERVICES_DIR}/.env")
+
+dotenv.load_dotenv(dotenv_path=f"{ROOT_DIR}/settings/.env")
```
fedorgrab, line 16: I'd still rename this settings to `conf`, to not to confuse it with this file. IMHO, totally personal naming preference. Up to you.

fedorgrab (approved):
Great stuff! I left some comments but they all are naming recommendations.

src/settings/sample_env
nmalfroy: I wanted to leave .env out of the name since gitignore then picks up the file

src/casp/services/_settings.py
nmalfroy: If it's ok, I'd like to leave it as settings since we call them settings in the code.  I think it would be confusing to have two names

src/casp/services/utils/gc_utils.py
```diff
@@ -1,16 +1,37 @@
...
-    credentials = Credentials.from_service_account_info(
-        info=settings.GOOGLE_ACCOUNT_CREDENTIALS, scopes=None, default_scopes=None
-    )
-    return credentials, settings.GOOGLE_ACCOUNT_CREDENTIALS.get("project_id")
+logger = logging.getLogger(__name__)
+
+
+def get_google_service_credentials() -> t.Tuple[Credentials, str]:
+    if settings.GOOGLE_ACCOUNT_CREDENTIALS != {}:
+        logger.warn(
```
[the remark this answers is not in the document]
nmalfroy: ooooh weird.  Yep, great call. I'll updated to warning

.github/workflows/docker-workflow.yml
```diff
@@ -0,0 +1,75 @@
...
+        with:
+          docker-registry-name: ${{ env.DOCKER_REGISTRY_NAME }}
+          image-name: ${{ env.PYTORCH_IMAGE_NAME }}
+          image-tag: ${{ env.IMAGE_TAG }}
+          image-type: ${{ matrix.image-type }}
+          gcp-provider-id: ${{ secrets.GCP_PROVIDER_ID }}
+          gcp-service-account-email: ${{ secrets.GCP_SERVICE_ACCOUNT_EMAIL }}
+          add-latest-tag: ${{ inputs.add-latest-tag }}
+
+      - id: build-cudo
```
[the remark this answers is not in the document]
nmalfroy: gah, yep

PR #154, opened 2024-05-20
Make sure that sentry context is set before router

nmalfroy:
The trick was setting the sentry context _before_ invoking the `call_next` method (which is where you eventually get into the router)

This also adds decoration to the error responses to ensure that we always return the x-trace-id header even when there's an error

### DIFF
--- a/src/casp/services/api/exception_handlers.py
+++ b/src/casp/services/api/exception_handlers.py
@@ -1,8 +1,22 @@
 from fastapi import Request, responses, status
+from starlette_context import context
 
 from casp.services._auth.exceptions import TokenException
 from casp.services.api.data_manager.exceptions import NotFound
 from casp.services.api.services.exceptions import APIBaseException
+from casp.services.constants import ContextKeys, HeaderKeys
+
+
+def decorate_response_headers(init_headers: dict[str, str] = None):
+    """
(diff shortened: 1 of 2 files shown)


KevinCLydon (approved):

src/casp/services/api/exception_handlers.py
```diff
@@ -1,8 +1,22 @@
 from fastapi import Request, responses, status
+from starlette_context import context
 
 from casp.services._auth.exceptions import TokenException
 from casp.services.api.data_manager.exceptions import NotFound
 from casp.services.api.services.exceptions import APIBaseException
+from casp.services.constants import ContextKeys, HeaderKeys
+
+
+def decorate_response_headers(init_headers: dict[str, str] = None):
```
fedorgrab, line 10: I think we should eventually start using "list", "dict", instead of "t.Dict", "t.List" as they got outdated after 3.10. But since we use it everywhere, shouldn't we keep doing them until we change them all at once? If not, then maybe I should start using newer annotations as well?

fedorgrab (approved):

src/casp/services/api/exception_handlers.py
nmalfroy: Maybe I'll keep this and do a followon to convert all instances to the more native `list` and `dict` if that's ok

PR #157, opened 2024-05-28
CAS-25 Deploy CAS from GitHub Action

nmalfroy:
This PR adds a brand new GitHub workflow + action that will let us deploy the built docker image to CAS services via github.

Once merged, it will be doable from the UI but until then, this can be run via the github CLI using:

```
    GIT_REF=<branch, git commit, tag, etc.> # Git reference to deploy the image from (helpful if modifying the deploy scripts or configuration files)
    IMAGE_TAG=<docker tag> # Tag to deploy the docker image from.
    SERVICE_ACCOUNT=sa-user@<project>.iam.gserviceaccount.com # Service account that will be running the service
    SECRET_REF=secret-name:latest # Reference to secret in the project's google secret manager as <secret name>:<version or latest> (note that the service account must have access to the secret)
    DB_CONNECTION=example-project:us-region-example:db-cluster-name # Cloud SQL connection name
    VPC_CONNECTOR=projects/<project>/locations/<region>/connectors/<connector> # VPC connector to use for the service
    DEPLOYMENT_PREFIX=example # Prefix to use for the deployment names
    DEPLOYMENT_FLAVOR=default # Flavor of the deployment to use. Must match a config file in the casp/services/deploy/configs directory

    gh workflow run deploy-workflow.yml repo cellarium-ai/cellarium-cloud \
            --ref $GIT_REF \
            -f image-tag=$IMAGE_TAG \
            -f service-account-email=$SERVICE_ACCOUNT \
            -f sql-instance=$DB_CONNECTION \
            -f vpc-connector=$VPC_CONNECTOR \
            -f config-secret=$SECRET_REF \
            -f deployment-prefix=$DEPLOYMENT_PREFIX \
            -f flavor=$DEPLOYMENT_FLAVOR
```

This will deploy the 3 CAS services and admin and model will be behind the VPC.

**Note: For this to work: you must remove the `MODEL_SERVER_URL` parameter in the secret you are using to deploy.**

This will deploy services named `<prefix>-[cas-model|cas-admin|cas-api]` with labels that include the prefix and the github user that triggered the deploy:

image

The GitHub run will output the URLs to the deployments:
image

A couple of other drive-bys that I also did as part of the PR for things that came up while debugging:

- The version of uvicorn is pinned to 0.29.0 to avoid log chatter and a potentially bigger issue with background thread management
- I added a locator field to the stackdriver logs to make it easier to tell where random log messages from from.  It looks like:

image

### DIFF
--- a/.github/actions/docker-deploy/action.yml
+++ b/.github/actions/docker-deploy/action.yml
@@ -30,9 +30,11 @@ inputs:
   deployment-project:
     required: true
     description: The project to deploy the image to
-  deployment-service-account:
+  deployment-service-account-email:
     required: true
-    description: The service account used as the identity for the deployment
+    description: |
+      The service account used as the identity for the deployment.  This is the account that must be granted access to Google cloud
+      resources that CAS services will be using (e.g. buckets)
   deployment-sql-instance:
     required: true
     description: The SQL instance to connect to in the form [project]:[region]:[instance]
@@ -152,7 +154,7 @@ runs:
           --min-instances ${{ env.MIN_INSTANCES }}
           --concurrency ${{ env.CONCURRENCY }}
           --port 8000
-          --service-account ${{ inputs.deployment-service-account }}
+          --service-account ${{ inputs.deployment-service-account-email }}
           --ingress ${{ env.INGRESS }}
           --command ${{ env.COMMAMD }}
           --args "${{ env.ARGS }}"
--- a/.github/actions/docker-deploy/action.yml
+++ b/.github/actions/docker-deploy/action.yml
@@ -134,6 +134,9 @@ runs:
           /app/settings/.env=${{ inputs.deployment-config-secret }}
         secrets_update_strategy: overwrite
         skip_default_labels: true
+        labels: |
+          deployment-prefix=${{ inputs.deployment-prefix }}
+          deployed-by=${{ github.actor }}
         env_vars: |
           MODEL_SERVER_URL=${{ env.MODEL_SERVER_URL }}
         flags: |
--- a/.github/actions/docker-deploy/action.yml
+++ b/.github/actions/docker-deploy/action.yml
@@ -1,0 +1,169 @@
+name: Deploy Docker Image to Cloud Run
+description: Deploy a docker image using Cloud Run.  This will override any deployment with the same prefix.  Returns the URL of the deployed image as `deployed-url`.
+inputs:
+  docker-registry-name:
+    required: true
+    description: GCP Docker Registry name
+  gcp-provider-id:
+    required: true
+    description: CAS GCP Provider ID
+  gcp-service-account-email:
+    required: true
+    description: CAS Deployer Service Account email
+  image-name:
+    required: true
+    description: Name of the image to deploy
+  image-tag:
+    required: true
+    description: Tag of the image to deploy
+  deployment-type:
+    required: true
+    description: The type of deployment to create (cas-model, cas-admin, cas-api)
+  deployment-prefix:
+    required: true
+    description: The prefix to add to the deployment name
+  deployment-flavor:
+    required: true
+    description: |
+      The nature of the deployment set.  E.g. is is a default or large deployment.  Configurations are set based on this value and
+      the values can be configured in the src/casp/services/deploy/cloudrun/{flavor}.json files.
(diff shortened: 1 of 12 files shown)


.github/actions/docker-deploy/action.yml
```diff
@@ -0,0 +1,172 @@
+name: Deploy Docker Image to Cloud Run
+description: Deploy a docker image using Cloud Run.  This will override any deployment with the same prefix.  Returns the URL of the deployed image as `deployed-url`.
+inputs:
+  docker-registry-name:
+    required: true
+    description: GCP Docker Registry name
+  gcp-provider-id:
+    required: true
+    description: CAS GCP Provider ID
+  gcp-service-account-email:
```
KevinCLydon, line 10: This is an extreme nitpick, but it's weird to me that this is "service account email" and deployment-service-account is just "service account".  I also think that the names+descriptions for those two inputs don't super clearly explain the difference between them.

.github/workflows/deploy-workflow.yml
```diff
@@ -0,0 +1,117 @@
...
+      config-secret:
+        description: 'The Google secret to use for the deployment configuration in the form [secret-name]:[version (or "latest")]'
+        required: true
+        default: ''
+      region:
+        description: 'The region to deploy the image to'
+        required: false
+        default: 'us-central1'
+      flavor:
+        description: 'The nature of the deployment set.  E.g. is is a default or large deployment.  Configurations are set based on this value and the values can be configured in the src/casp/services/deploy/cloudrun/{flavor}.json files.'
```
KevinCLydon, line 34: "is is"

src/casp/services/logging.py
```diff
@@ -92,12 +92,21 @@ def format(self, record):
                 global_log_fields["logging.googleapis.com/trace"] = trace.get_trace(self.project_id)
                 global_log_fields["logging.googleapis.com/spanId"] = trace.span_id
 
+            locator = {
+                "filename": record.filename,
+                "lineno": record.lineno,
+                "funcname": record.funcName,
```
KevinCLydon, line 98: Is this actually the variable name python logging records use?  "funcName" doesn't seem pythonic

src/casp/services/deploy/cloudrun/README.rst
```diff
@@ -0,0 +1,34 @@
...
+==================
+
+This directory should contain various deployment configurations for the sizes of services used by the deployment GitHub action.
+These are JSON files named `[flavor].json` and the contents should look like:
+
+.. code-block:: json
+
+    {
+      "cas-admin": {
+        "cpu": "1000m",
```
KevinCLydon, line 11: Probably a dumb question: what is a "1000m" CPU?

KevinCLydon (changes requested):
As usual, only a couple very small things.  Otherwise, this looks awesome!

.github/actions/docker-deploy/action.yml
nmalfroy: Fait enough!  Lemme take another stab at it

src/casp/services/logging.py
nmalfroy: It sure is...

nmalfroy: Yeah, it's weird that such a core module would have this one random different casing but yeah, double checked:
https://github.com/python/cpython/blob/e69d068ad0bd6a25434ea476a647b635da4d82bb/Doc/library/logging.rst#logrecord-attributes

src/casp/services/deploy/cloudrun/README.rst
nmalfroy: Not dumb at all!  1000m CPU is the same as 1 CPU.  Often for shared resources, CPU is measured in sub-CPU units (e.g. milli-CPUs in this case)

KevinCLydon: Oh that's really interesting!  I did not know that.  Thanks for explaining!

KevinCLydon (approved):

fedorgrab (approved):
Looks good to me. Honestly, I don't know much about workflows, so I don't have much feedback to provide. I'm just learning from this PR. It looks great, and it's a great feature!

PR #159, opened 2024-06-05
Fix: DB multiple head revision

fedorgrab:
After merging our previous PRs, our migrations had two "heads," which was caused by having the same reference in 2 latest migrations, preventing automatic migration application. This PR fixes it.


nmalfroy (approved):
Good catch!

KevinCLydon (approved):

PR #164, opened 2024-06-21
Created a workflow for spinning up a test server and running integration tests against it

KevinCLydon:
Quick summary of the process:

1. Builds the docker image for cellarium-cloud for the specified version
2. Deploys the model service
3. Deploys the API service
4. If user set `deploy-index` to `true`, deploys the specified vector search index
5. Checks out the specified version of cellarium-cas and runs integration tests (currently just the one that was already there)
6. If user set `deploy-index` to `true`, undeploys the specified vector search index

A couple points that warrant further discussion for this PR:

1. When you run the gcloud commands to deploy/undeploy an index, it kicks off the operation and returns immediately.  For the deploy operation, it gives you an operation ID that I am checking repeatedly so it stays on that step until the deploy is complete.  However, undeploy doesn't work this way.  It doesn't return an operation ID, and I can't find a way to look one up.  For now, I don't have any wait logic for undeploy because I figure it's not strictly necessary for the action to complete successfully.  If we think it's necessary, I can maybe do something with reading the log messages from repeated attempts to undeploy, since there are different ones for "that index is currently being undeployed" and "that index is not deployed".
2. Because of the limit on Github Actions workflow inputs, I'm currently hardcoding the ID and display name of the deployed index to `integration_test_deployed_index`.  Theoretically, this is fine as long as you don't have an existing deployed index with that name or ID.  I'm not sure if we want to go another route like providing a config json to allow specifying those things on run.  However, this ties in to another possible complication / discussion point
3. The index you deploy isn't going to be dynamically set to be the one that we test against.  The info for that deployed index needs to already exist in the test DB.  You're basically just inputting the values that match your default index in the test DB.  We could theoretically make it work dynamically if we deployed the index and then wrote its info to the specified test DB, but that does add a level of complexity to this workflow that maybe wouldn't really justify itself with added value.  I'm undecided.

### DIFF
--- a/.github/actions/integration-tests/action.yml
+++ b/.github/actions/integration-tests/action.yml
@@ -1,0 +1,42 @@
+name: Run cellarium-cas integration tests
+description: Runs the integration tests from the cellarium-cas repository
+inputs:
+  revision:
+    description: The revision of the cellarium-cas repository to use
+    required: false
+    default: main
+  api-key:
+    description: The API key for connecting to the CAS API
+    required: true
+  api-url:
+    description: The URL of the CAS API service
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - id: checkout
+      name: Checkout
+      uses: actions/checkout@v4
+      with:
+        repository: "cellarium-ai/cellarium-cas"
+        ref: ${{ inputs.revision }}
+        path: "./cellarium-cas"
+    
+    - id: install-dependencies
+      name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements/dev.txt
+      shell: bash
+      working-directory: ./cellarium-cas
+    
+    - id: run-tests
+      name: Run tests
+      env:
+        TEST_API_TOKEN: ${{ inputs.api-key }}
+        TEST_API_URL: ${{ inputs.api-url }}
+      run: |
+        tox -e integration 
+      shell: bash
+      working-directory: ./cellarium-cas
--- a/.github/actions/vector-search-index-deploy/action.yml
+++ b/.github/actions/vector-search-index-deploy/action.yml
@@ -1,0 +1,74 @@
+name: Deploy vector search index
+description: Creates a deployment of the specified vector search index
+inputs:
+  gcp-provider-id:
+    required: true
+    description: CAS GCP Provider ID
+  gcp-service-account-email:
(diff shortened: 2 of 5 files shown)


.github/actions/integration-tests/action.yml
```diff
@@ -0,0 +1,42 @@
+name: Run cellarium-cas integration tests
+description: Runs the integration tests from the cellarium-cas repository
+inputs:
+  revision:
+    description: The revision of the cellarium-cas repository to use
```
nmalfroy, line 5: This _may_ change once we're publishing to pypi

.github/actions/vector-search-index-deploy/action.yml
```diff
@@ -0,0 +1,74 @@
...
+      uses: actions/checkout@v4
+
+    - id: google-login
+      name: Authenticate with Google Cloud
+      uses: google-github-actions/auth@v2
+      with:
+        token_format: access_token
+        workload_identity_provider: ${{ inputs.gcp-provider-id }}
+        service_account: ${{ inputs.gcp-service-account-email }}
+        access_token_lifetime: 1500s
```
nmalfroy, line 43: Might be a little easier to read in minutes?

nmalfroy (approved):
Nothing really to add that I can see.  This looks great!

.github/actions/integration-tests/action.yml
KevinCLydon: Oh really?  What will change?

nmalfroy: That we might want to pip install from `cellarium-cas` rather than a git hash

.github/actions/vector-search-index-deploy/action.yml
nmalfroy: oh gotcha.  Weird that it would still require an `s` at the end of the number then

fedorgrab (approved):
Great stuff!

.github/actions/integration-tests/action.yml
KevinCLydon: Would that prevent us from running the integration tests for arbitrary git hashes of the client repo?  I think we want to have the ability to do that

nmalfroy: Yeah, that's a fantastic point.  I retract my comment :)

PR #166, opened 2024-06-27
Added description fields to model and index

KevinCLydon:
Now you can describe models and indexes.

I just kind of picked an arbitrary size for the strings in the DB.

### DIFF
--- a/src/casp/services/api/schemas/cellarium_general.py
+++ b/src/casp/services/api/schemas/cellarium_general.py
@@ -1,15 +1,19 @@
 import datetime
-from typing import Optional
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, validator
 
 
 class CASModel(BaseModel):
     model_name: str = Field(example="human-pca-001")
-    description: Optional[str] = Field(example="1st version of PCA model for human data")
+    description: str = Field(example="1st version of PCA model for human data", default="")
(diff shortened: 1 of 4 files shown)


fedorgrab (changes requested):
It seems like you fgt to include migration file, or am I missing smth?

KevinCLydon:
> It seems like you fgt to include migration file, or am I missing smth?

Oh you're right!  I forgot to add them to the git repo.  I'll fix that

fedorgrab (approved):
Cool stuff!

src/casp/services/api/schemas/cellarium_general.py
```diff
@@ -1,11 +1,12 @@
 import datetime
+from typing import Optional
 
 from pydantic import BaseModel, Field
 
 
 class CASModel(BaseModel):
     model_name: str = Field(example="human-pca-001")
-    description: str = Field(example="1st version of PCA model for human data")
+    description: Optional[str] = Field(example="1st version of PCA model for human data")
```
nmalfroy, line 9: Conversely, we could also make sure to always return a `""` if it's empty - it simplifies handling on the client side I think.  Like, if there's an null, I think it would end up printing "None" in the description output I think

nmalfroy (approved):
Nice, looks good!

PR #170, opened 2024-07-02
CAS-44 Update default email sender and update email with correct pip install

nmalfroy:
What it says on the tin.  I updated the config in prod but I wanted to have the default reflect the working value.  Also updates the pip install to not install from github


KevinCLydon (approved):
Looks good to me 👍

PR #171, opened 2024-07-03
Added an endpoint for checking the client version number for compatibility

KevinCLydon:


src/casp/services/_settings.py
```diff
@@ -34,6 +34,7 @@ class AllEnvSettings(BaseSettings):
     LOG_LEVEL: str = "info"
     LOG_CONFIG: str = f"{SERVICES_DIR}/log_config.yaml"
     LOG_AS_JSON: bool = os.environ.get("LOG_AS_JSON", True)
+    MIN_CLIENT_VERSION: str = "1.4.0"
```
nmalfroy, line 37: We probably actually want this to be a constant since it's linked to the code (I don't think it would make sense to update this via config right?)

src/casp/services/api/services/cellarium_general_service.py
```diff
@@ -23,6 +26,18 @@ def get_application_info(self) -> schemas.ApplicationInfo:
         """
         return self.cellarium_general_dm.get_application_info()
 
+    def validate_client_version(self, client_version: str):
```
nmalfroy, line 29: I think we want a positive and negative unit test for this method at least

nmalfroy (changes requested):
Looks great!  Couple minor things

src/casp/services/_settings.py
KevinCLydon: That's a good point.  I'll make that change

src/casp/services/api/services/cellarium_general_service.py
KevinCLydon: Ooo good idea

fedorgrab (approved):

src/casp/services/api/services/exceptions.py
```diff
@@ -29,3 +31,20 @@ class CellMetadataColumnDoesNotExist(InvalidInputError):
 
 class QuotaExceededException(APIBaseException):
     http_code: int = status.HTTP_429_TOO_MANY_REQUESTS
+
+
+class InvalidClientVersionException(APIBaseException):
```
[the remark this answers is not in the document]
KevinCLydon: The body of the response from the server is just the error message, right?

nmalfroy (approved):
lgtm!

fedorgrab (approved):
Looks good to me!

PR #172, opened 2024-07-17
CAS-47 Merge the model service into the api service

nmalfroy:
This PR aims to improve performance of annotation calls by merging the model service into the API service.  This saves us having to:
- stream large amounts of data from one service to another and 
- serialize / deserialize responses

The second commit also changes the service code to pass a model object around rather than constantly requerying it from the DB.  This should take load off of the DB too.  Part of that effort also involved refactoring unit tests a little and creating a new set of unit tests for the authorization code (as a result of moving it into it's own module)


KevinCLydon (approved):
This looks awesome!  I only have a couple very minor things

src/casp/services/api/data_manager/cellarium_general.py
```diff
@@ -104,39 +105,28 @@ def get_model_by_name(self, model_name: str) -> models.CASModel:
         :return: CAS ML model object from the database
         """
         with self.system_data_db_session_maker() as session:
-            model = session.query(models.CASModel).filter_by(model_name=model_name).first()
+            model = (
+                session.query(models.CASModel)
+                # Load the matching engine eagerly to be able to access outside of a session
+                .options(orm.joinedload(models.CASModel.cas_matching_engine))
```
fedorgrab, line 111: That's a cool fix!

src/casp/services/api/services/authorization.py
```diff
@@ -0,0 +1,34 @@
+import typing as t
+
+from casp.services.api.data_manager import CellariumGeneralDataManager
+from casp.services.api.data_manager import exceptions as dm_exc
+from casp.services.api.services import exceptions
+from casp.services.db import models
+
+
+class Authorizer:
```
fedorgrab, line 9: Why make this a separate class? It has only one method.

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -139,17 +114,17 @@ def __validate_knn_response(embeddings: np.array, knn_response: MatchResult) ->
...
         :param embeddings: Embeddings to match.
-        :param model_name: Model name to use for matching.
+        :param model: Model to use for matching.
 
         :return: List of lists of MatchNeighbor objects.
         """
-        index = self.__get_match_index_endpoint_for_model(model_name=model_name)
+        index = model.cas_matching_engine
 
         matching_client = MatchingClient.from_index(index)
```
fedorgrab, line 129: Very small thing, just saving one line, but we can do this as well (and thus removing `index = model.cas_matching_engine`):
```suggestion
        matching_client = MatchingClient.from_index(model.cas_matching_engine)
```

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -383,23 +359,22 @@ async def annotate_cell_type_ontology_aware_strategy_with_activity_logging(
                 logger.error(f"Failed to log user activity: {e2}")
             raise e
 
-    async def annotate_cell_type_ontology_aware_strategy(
-        self, user: models.User, file: t.BinaryIO, model_name: str, prune_threshold: float, weighting_prefactor: float
+    async def __annotate_cell_type_ontology_aware_strategy(
```
fedorgrab, line 362: Why do you want to make it private? I think this method should be public based on what it does. Or maybe I am not seeing something, could you please explain?

fedorgrab (commented):
I left some comments, please take a look.

src/casp/services/api/services/authorization.py
nmalfroy: I don't know if that's a reason not to have it in there.  Logically it seems to make more sense in here than in the cell operations service so I wanted to pull it out. I originally pulled it out because I thought we were doing a similar thing in the general ops service and while we're not currently, it seems like we might and it makes sense (IMO) to have a place for authz checks like this to live.

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -299,9 +275,8 @@ async def annotate_cell_type_summary_statistics_strategy_with_activity_logging(
 
     async def annotate_cell_type_summary_statistics_strategy(
```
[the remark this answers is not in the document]
nmalfroy: Ah right given that it's private

src/casp/services/api/services/cell_operations_service.py
nmalfroy: Because this doesn't get (and probably should be) called from outside the class.  This just gets called from `annotate_cell_type_ontology_aware_strategy_with_activity_logging` which is the only thing that should be calling this I think.  Does that make sense?

fedorgrab: Oh, I forgot about those methods. Got it, makes sense then.

fedorgrab (approved):

PR #176, opened 2024-07-25
CAS-58 Fix Sentry Error Spam

nmalfroy:
This PR makes us create DB connections as needed as opposed to eagerly.  Managing connections eagerly seems to cause credential expiration errors - they are innocuous since the connection gets recreated but it spams our logs.  I've tested on my environment and it seems to make the exceptions disappear.


KevinCLydon (approved):
Looks good to me!

PR #177, opened 2024-07-26
CAS-55 Caching model files and cellarium modules so we don't keep downloading them

KevinCLydon:


src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -41,6 +41,14 @@ async def annotate(
     )
 
 
+@cell_operations_router.get("/cache-info")
+def get_cache_info():
```
nmalfroy, line 45: We could also leave this and make it admin only?

nmalfroy (approved):
Awesome!  And I can confirm that this helps some with perf.  Nicely done!

nmalfroy:
Oh and that unit test error is fixed in my PR (we need to pin the version of numpy to less than 2)

KevinCLydon:
> Oh and that unit test error is fixed in my PR (we need to pin the version of numpy to less than 2)

Oh that's good news.  Is there a breaking change in numpy 2 that cellarium-ml doesn't support or something?  I'll rebase this branch once your PR is merged

nmalfroy:
Should be good to merge with development branch now!

src/casp/services/api/routers/cell_operations_router.py
```diff
@@ -103,6 +103,15 @@ async def annotate_cell_type_ontology_aware_strategy(
     )
 
 
+@cell_operations_router.get("/cache-info", response_model=schemas.CacheInfo)
+def get_cache_info(user: models.User = Depends(dependencies.authenticate_user)):
```
fedorgrab, line 107: Why do we need this method exposed? I thought it was intended to be removed before merging.

KevinCLydon: @nmalfroy proposed we could keep it as admin-only.  I do think it could be useful for diagnosing issues, so I cleaned it up and added that admin-only check to it.  Thoughts?

nmalfroy: Yeah, I like the idea of keeping it as an API only (e.g. not client) endpoint for admins to verify the cache

fedorgrab (approved):

PR #179, opened 2024-08-07
CAS-57 Added a check to annotate methods to return an error if the anndata file is the wrong type

KevinCLydon:
Backend handling for anything not caught by cellarium-ai/cellarium-cas#78

### DIFF
--- a/src/casp/services/api/services/cell_operations_service.py
+++ b/src/casp/services/api/services/cell_operations_service.py
@@ -60,18 +60,27 @@ class CellOperationsService:
             module_cache_info=cache_info["module_cache_info"],
         )
 
-    def __validate_anndata_type(self, adata: anndata.AnnData):
+    def __validate_anndata_file(self, adata: anndata.AnnData):
         """
-        Validates that the anndata matrix is float32, which is a requirement for cellarium-ml
+        Validates the type of the anndata matrix and the presence of the 'total_mrna_umis' column,
+        which are required for the data to be processed properly by our cellarium-ml models.
 
         :param adata: Anndata object to validate
 
         :raises exceptions.InvalidAnndataFileException: If the anndata matrix is not float32
         """
         if adata.X.dtype != np.float32:
-            raise exceptions.InvalidAnndataFileException(adata.X.dtype)
-        if adata.obs["total_mrna_umis"].dtype != np.float32:
-            raise exceptions.InvalidAnndataFileException(adata.obs["total_mrna_umis"].dtype)
+            raise exceptions.InvalidAnndataFileException(
+                f"CAS only support anndata matrices with dtype 'float32'. The provided matrix has type '{adata.X.dtype}'."
+            )
+        if "total_mrna_umis" not in adata.obs:
+            raise exceptions.InvalidAnndataFileException(
+                "Provided matrix does not contain 'total_mrna_umis' column. This column is required for CAS to work."
+            )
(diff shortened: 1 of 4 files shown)


src/casp/services/api/services/cell_operations_service.py
```diff
@@ -60,36 +60,53 @@ def get_cache_info(user: models.User) -> schemas.CacheInfo:
...
-        :param file: Anndata file to get the number of cells from.
+        :param adata: Anndata object to validate
 
-        :return: Number of cells in the anndata file.
+        :raises exceptions.InvalidAnndataFileException: If the anndata matrix is not float32
+        """
+        if adata.X.dtype != np.float32:
+            raise exceptions.InvalidAnndataFileException(adata.X.dtype)
+        if adata.obs["total_mrna_umis"].dtype != np.float32:
+            raise exceptions.InvalidAnndataFileException(adata.obs["total_mrna_umis"].dtype)
```
fedorgrab, line 74: This will raise an error with message: 
`f"CAS only supports anndata matrices with dtype 'float32'. The provided matrix has dtype '{dtype}'"` 
Which seems confusing as an error message as it points to the matrix, but it is actually `obs["total_mrna_umis"]` which has to be fixed.

I think it's worth defining the error without a message in the constructor and pass a message when it's raised. So that you'd raise the error like this:
```
raise exceptions.InvalidAnndataFileException(f"CAS only supports anndata matrices with dtype 'float32'. The provided matrix has dtype '{dtype}'")
```
or 
```
f"Provided matrix's column `total_mrna_umis` should be with type of `np.float32`, got {adata.obs['total_mrna_umis']} instead"
```

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -60,36 +60,53 @@ def get_cache_info(user: models.User) -> schemas.CacheInfo:
...
+
+    def __read_anndata_file(self, file: t.BinaryIO) -> anndata.AnnData:
```
fedorgrab, line 76:
```suggestion
    def __read_and_validate_anndata_file(self, file: t.BinaryIO) -> anndata.AnnData:
```

src/casp/services/api/services/cell_operations_service.py
```diff
@@ -60,36 +60,53 @@ def get_cache_info(user: models.User) -> schemas.CacheInfo:
             module_cache_info=cache_info["module_cache_info"],
         )
 
-    def __get_cell_count_from_anndata(self, file: t.BinaryIO) -> int:
+    def __validate_anndata_type(self, adata: anndata.AnnData):
```
fedorgrab, line 63: I think it's a great call to have this function. I think it'd be also important to include the check for`total_mrna_umis` ( if it is in `adata["obs"]`). It's a computationally cheap check, but it's crucial to have it as without it, model_service would raise an error.

fedorgrab (changes requested):

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: Oh you're right.  I completely missed that.  I'll update it

src/casp/services/api/services/cell_operations_service.py
KevinCLydon: Ooo that's a good idea.  I'll add that

nmalfroy (approved):
This looks good to m!

fedorgrab (approved):
LGMT

PR #181, opened 2024-08-16
Add script building cell type ontology

fedorgrab:
* Created a `scripts` directory for all executable scripts. I suggest moving the `bq_scripts` to this directory as well in the future. 
* Renamed infrastructure tests to integration tests, as they are actually integration tests.
* Updated the Cell Type Ontology resource to match the one used in schema 5.
* Added scripts for building the Cell Type Ontology resource.

I recommend continuing to use ontology resources instead of calculating them during instance creation, as building the resource takes approximately 40-60 seconds, which would double the time required to create new instances.

I also propose that we build resource data classes in a future refactor (which we should discuss separately) once we have established a shared data classes library across the entire project.


nmalfroy (approved):
LGTM!

PR #183, opened 2024-09-20
CAS-74 Added a default lifetime cell quota for users

KevinCLydon:
When a user is created, they will have a default lifetime quota of 200k cells.  When they submit a request, we now also check their lifetime cells processed against their lifetime quota and send a quota exceeded error message if this request exceeds their remaining quota.


src/casp/services/api/data_manager/cellarium_general.py
```diff
@@ -143,16 +143,25 @@ def log_user_activity(
...
+        # because of any weirdness in the simultaneous processing of cells)
+        remaining_quota = min(
+            user.cell_quota - self.get_cells_processed_this_week_for_user(user),
+            (
+                user.lifetime_cell_quota - user.total_cells_processed
+                if user.lifetime_cell_quota is not None
+                else float("inf")
+            ),
+        )
+        return max(remaining_quota, 0)
```
fedorgrab, line 164: nit, but I think it could be more readable if you break it into steps like this:
```python3
weekly_remaining_quota = user.cell_quota - self.get_cells_processed_this_week_for_user(user=user)

lifetime_remaining_quota = (
    user.lifetime_cell_quota - user.total_cells_processed
    if user.lifetime_cell_quota is not None
    else float("inf")
)

remaining_quota = min(weekly_remaining_quota, lifetime_remaining_quota)

# Ensure the remaining quota is non-negative
return max(remaining_quota, 0)
```

src/casp/services/api/schemas/cellarium_general.py
```diff
@@ -1,4 +1,5 @@
 import datetime
+from typing import Optional
```
fedorgrab, line 2: Another nit, but `import typing as t` would follow how we do it in other files.

src/casp/services/api/services/cellarium_general_service.py
```diff
@@ -94,7 +94,11 @@ def get_quota_for_user(self, user: models.User) -> schemas.UserQuota:
 
         return schemas.UserQuota(
             user_id=user.id,
-            quota=user.cell_quota,
-            remaining_quota=self.cellarium_general_dm.get_remaining_quota_for_user(user=user),
+            weekly_quota=user.cell_quota,
+            remaining_weekly_quota=self.cellarium_general_dm.get_remaining_quota_for_user(user=user),
             quota_reset_date=quota_reset_date,
+            lifetime_quota=user.lifetime_cell_quota,
+            remaining_lifetime_quota=(
```
fedorgrab, line 101: nit: Consider factoring out `None if user.lifetime_cell_quota is None else user.lifetime_cell_quota - user.total_cells_processed` into a separate variable. This could improve readability. For example:

```python3
remaining_lifetime_quota = (
    user.lifetime_cell_quota - user.total_cells_processed
    if user.lifetime_cell_quota is not None
    else None
)
```

src/casp/services/api/data_manager/cellarium_general.py
fedorgrab: Just a thought: it might make more sense to move the quota calculation into a service method rather than keeping it in the data manager. This could better align with the separation of concerns principle.

fedorgrab (changes requested):
I left some comments for discussion, mostly small nits. There's only one thought about moving quota logic to a service class, let me know your thoughts about it!

src/casp/services/api/data_manager/cellarium_general.py
KevinCLydon: I think I had originally started doing it this way when I was writing the original quota code, but I realized both the cell operations service and the cellarium general service need to use this method, so I wasn't sure if it made sense to move it into one or the other.  I guess maybe we could make a new file to just put the quota stuff in?  What do you think?

fedorgrab: I like this idea. In the future, it will only help us better organize and navigate the code!
I think it might make sense to unify this into another "service" class, like a quota service or something similar.

nmalfroy (approved):
This looks great!

src/casp/services/api/data_manager/cell_quota.py
```diff
@@ -0,0 +1,65 @@
...
+class CellQuotaDataManager(BaseDataManager):
+
+    # SQL query directory
+    SQL_QUERY_DIR = f"{settings.SERVICES_DIR}/api/data_manager/sql_queries"
+
+    # SQL queries from files
+    with open(f"{SQL_QUERY_DIR}/get_cells_processed_this_week_for_user.sql.sa", "r") as f:
+        SQL_GET_CELLS_PROCESSED_THIS_WEEK = f.read()
+
+    def get_remaining_quota_for_user(self, user: models.User) -> int:
```
fedorgrab, line 19: What do you think about placing this method under the CellQuotaService, since it is responsible for quota calculation rather than retrieving information from the database?

fedorgrab (changes requested):

src/casp/services/api/data_manager/cell_quota.py
KevinCLydon: I tried that initially, but got a not-fully-loaded-module/circular-import error.  I assume it was because `CellOperationsService` calls that method, so I had to import `CellQuotaService` in `CellOperationsService` and then both are also imported by the `services` module?  Although I wonder if maybe the import order caused the issue?  I'm not sure.  Any thoughts?

src/casp/services/api/services/cell_quota_service.py
```diff
@@ -0,0 +1,68 @@
...
+        return schemas.UserQuota(
+            user_id=user.id,
+            weekly_quota=user.cell_quota,
+            remaining_weekly_quota=self.get_remaining_quota_for_user(user=user),
+            quota_reset_date=quota_reset_date,
+            lifetime_quota=user.lifetime_cell_quota,
+            remaining_lifetime_quota=remaining_lifetime_quota,
+        )
+
+    def get_remaining_quota_for_user(self, user: models.User) -> int:
```
fedorgrab, line 45: Sorry, but could you please move this above `get_quota_for_user` to maintain the order of methods that execute one another?

fedorgrab (changes requested):

src/casp/services/api/services/cell_quota_service.py
KevinCLydon: Yup, I can do that

fedorgrab (approved):
LGTM!

PR #189, opened 2024-10-22
Fixed dependency issue causing "Create" pages on the admin dashboard to break

KevinCLydon:
A dependency of a dependency had an update that appears to be broken, so now we'll explicitly specify the version of that dependency to use.


fedorgrab:
Closing this one as it's the same as #188

PR #191, opened 2024-10-24
Now quotas start lower but they can be increased via the REST API

KevinCLydon:


src/casp/services/api/data_manager/cell_quota.py
```diff
@@ -38,3 +38,20 @@ def get_cells_processed_this_week_for_user(self, user: models.User) -> int:
...
             return session.execute(cell_count_sum_query, {"id": user.id, "start_of_week": start_of_week}).first()[0]
+
+    def increase_quota_for_user(self, user: models.User, new_quota: int):
+        """
+        Increase cell processing quota for a specific user
+
+        :param user: User object to increase quota for
+        :param new_quota: The new quota to set for the user
+
+        :return: User object with updated quota
```
fedorgrab, line 49: Seems like this method doesn't return anything. I guess we don't need description for return?

But I'd put `-> None` in the method return annotation like this:
`def increase_quota_for_user(self, user: models.User, new_quota: int) -> None:`

src/casp/services/api/data_manager/cellarium_general.py
```diff
@@ -110,6 +110,20 @@ def get_model_by_name(self, model_name: str) -> models.CASModel:
...
+    def get_user_by_email(self, user_email: str) -> models.User:
+        """
+        Get user by email
+
+        :param user_email: Email of the user to get
+
+        :return: User object
+        """
+        with self.system_data_db_session_maker() as session:
+            user = session.query(models.User).filter(models.User.email == user_email).one_or_none()
```
fedorgrab, line 122: Wow, never saw `one_or_none` that seems handy!

src/casp/services/api/services/cell_quota_service.py
```diff
@@ -65,4 +67,25 @@ def get_quota_for_user(self, user: models.User) -> schemas.UserQuota:
...
+
+        :param admin_user: User object of the admin user requesting the quota increase
+        :param user_for_increase: User object to increase quota for
+
+        :return: None
+
+        :raises: AccessDeniedError if admin_user is not an admin
+        """
+        if not admin_user.is_admin:
+            raise exceptions.AccessDeniedError("Access denied.  This is an admin-only endpoint.")
```
fedorgrab, line 86: Seems like you have 2 spaces in the messages description in "...denied.  This..."

src/casp/services/db/migrations/versions/t_2024-10-23_13-39-07_quota_increase.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "cdfe110d1068"
+down_revision = "c80bd73203d1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("users_user", sa.Column("quota_increased", sa.Boolean(), nullable=False, server_default="False"))
```
fedorgrab, line 20: I think that if you ran this migration in the db where users exist, it'd raise error, because the field is not nullable. I had this issue, but I didn't use server_default. Still, my understanding is that it is not going to apply default on fly to all existing entities. Please, correct me if I'm wrong if you tested this particular case. 
Otherwise, you can take a look at migration`37fbe8997eca_cas_model_is_default.py` there's a workaround solution for this.

fedorgrab: Also, I believe since we're using postgres it'd be better to put "false" as a server_default. Not sure 100% tho, but my understanding is that this is going directly into the database and I am not sure that postgres would accept 'False' (capitalized). This is what their docs say about their booleans:
https://www.postgresql.org/docs/6.3/c0805.htm

fedorgrab (changes requested):

src/casp/services/db/migrations/versions/t_2024-10-23_13-39-07_quota_increase.py
KevinCLydon: I can confirm this works.  Using "server_default" is the solution I found to the error you described.  Also, postgres treats booleans as case-insensitive.  I think maybe all keywords are case-insensitive?

src/casp/services/api/services/cell_quota_service.py
KevinCLydon: Oops, that's an old habit

src/casp/services/api/data_manager/cellarium_general.py
KevinCLydon: It is!

src/casp/services/api/data_manager/cell_quota.py
KevinCLydon: Oh oops.  I originally had it returning the user object, but changed my mind.  I'll fix that

src/casp/services/db/migrations/versions/t_2024-10-23_13-39-07_quota_increase.py
fedorgrab: Oh, that's a neat solution then! About case-insensitivity - cool, then I am resolving this message.

mbabadi: This conversation got very esoteric all of a sudden. I see myself out.

src/casp/services/db/migrations/versions/t_2024-10-23_13-39-07_quota_increase.py
```diff
@@ -0,0 +1,25 @@
...
+    op.execute("update users_user set quota_increased = true where lifetime_cell_quota = 200000")
```
mbabadi, line 21: Please use `INCREASED_LIFETIME_QUOTA` from `_settings`.

mbabadi: Oh wait-- ignore if migration is automatically generated code.

mbabadi (approved):

src/casp/services/db/migrations/versions/t_2024-10-23_13-39-07_quota_increase.py
KevinCLydon: It's partially autogenerated, but that particular line is not.  I'll make that change

src/casp/services/api/services/cell_quota_service.py
```diff
@@ -65,4 +67,25 @@ def get_quota_for_user(self, user: models.User) -> schemas.UserQuota:
             quota_reset_date=quota_reset_date,
             lifetime_quota=user.lifetime_cell_quota,
             remaining_lifetime_quota=remaining_lifetime_quota,
+            quota_increased=user.quota_increased,
         )
+
+    def increase_quota(self, admin_user: models.User, user_for_increase: models.User):
```
fedorgrab, line 73: Could you please add `-> None` in the method definition and remove :return: None from the docstring here as well?

fedorgrab (changes requested):
Just one tiny nit.

fedorgrab (approved):

PR #196, opened 2024-11-13
Fix for the timestamp bug

KevinCLydon:
For datetime columns on the user and user activity tables, we were setting the default to the result of calling datetime.now() once.  So it was basically only calling it the first time it checked for the default value and then only ever calling it again in new instances of the server (I think).  This resulted in a bunch of incorrect datetime values.  I've fixed it by replacing it with a lambda that calls datetime.now() each time a new user or user activity object is instantiated.


fedorgrab (approved):
LGTM. However, Github actions don't pass, not sure if it's related to this PR though.

Issue #129, opened 2024-02-20
Internal CAS benchmarking script


fedorgrab:
The benchmarking script should:
1.  Be automated and executable through pipelines
2. Log summary at the end (possibly to Neptune)

PR #199, opened 2024-11-22
Feat: #129 Scientific Accuracy Benchmarking Scirpts

fedorgrab:
This PR adds:
1. Scripts for benchmarking. The benchmarking scripts have two main purposes:
* Generate CAS outputs: This script uses CAS with a specified CAS token and CAS API URL to produce CAS responses in a specified directory.
* Calculate metrics: This script calculates metrics for each output, requiring the datasets used for generating the outputs as well as the ground truths.

2. Updated ontology-generating resources: For benchmarking, an additional resource for the ontology is needed. This resource includes a cache-like precomputed structure where each ontology label has its own precomputed n_hops data (e.g., descendants, ancestors, nodes, etc.).

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -20,7 +20,7 @@ Table of Contents
    modules/readme_modules/db_service
    modules/readme_modules/wdl_workflows
    modules/readme_modules/data_manager
+   modules/readme_modules/scripts
    modules/changelog
    modules/running_locally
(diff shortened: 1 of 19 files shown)


fedorgrab:
Closing this PR as it is stale and now benchamrking is implemented in the [client side repo](https://github.com/cellarium-ai/cellarium-cas/pull/128)

PR #210, opened 2026-04-03
feat: add CellInfoMetadata and OntologicalColumn models and admin view

fedorgrab:
### TLDR;

Now backend supports multiple ontologies. Client still can only annotate cell type, but backend is flexible now, we will have to add additional handlers for other columns later on, but the whole mechanic of storing multiple cell info tables + multiple ontologies per table allows us to have as many models, trained on different data extracts, and whichever columns we want. Once we are ready to process disease, we will have to add it via admin just like in this example I added cell type:
image

Replaces the hardcoded global `GCS_CELL_ONTOLOGY_RESOURCE_FILE` config constant with a DB-backed, per-model configuration, allowing different models to reference different ontology versions and SOMA DataFrames independently.

### New DB models

- **`CellInfoMetadata`** — holds a SOMA DataFrame URI and a human-readable name; shareable across multiple `CASModel` records
- **`OntologicalColumn`** — child of `CellInfoMetadata`; one row per annotated variable (e.g. `cell_type`, `disease`), each with its own GCS ontology resource URI and a free-text description field for version/provenance notes
- `CASModel.cell_metadata_uri` replaced by `CASModel.cell_info_metadata` FK

### Migration (`t_2026-04-03`)

Creates the new tables, migrates existing `cell_metadata_uri` values into `CellInfoMetadata` records, and drops the old column. Fully reversible.

### Admin

- New **CellInfoMetadata** view under "ML Management" with `OntologicalColumn` inline editing
- `CASModel` form updated to select a `CellInfoMetadata` via dropdown

### Service & strategy

- `CellOntologyResource` now requires an explicit `ontology_resource_uri` parameter — no global fallback
- Service layer looks up the `cell_type` ontological column from the model's linked metadata and raises a clear error if not configured
- `get_model_by_name` eager-loads `cell_info_metadata` and `ontological_columns` to avoid detached-instance errors outside the session


cellarium/cas_backend/apps/admin/views.py
```diff
@@ -479,6 +501,9 @@ class CellIngestInfoAdminView(CellariumCloudAdminModelView):
     index_view=CellariumCloudAdminIndexView(url="/", template="admin/main_page.html"),
 )
 admin.add_view(UserAdminView(models.User, db_session, name="User", category="Users"))
+admin.add_view(
+    CellInfoMetadataAdminView(models.CellInfoMetadata, db_session, name="CellInfoMetadata", category="ML Management")
```
sjfleming: sonarcloud has some suggestion here

cellarium/cas_backend/apps/compute/services/cell_operations_service.py
```diff
@@ -369,9 +370,18 @@ async def __annotate_cell_type_ontology_aware_strategy(
         query_ids, knn_response = await self.get_knn_matches(adata=adata, model=model)
 
         logger.info("Applying CellTypeOntologyAwareConsensusStrategy to the query results")
+        if not any(col.column_name == "cell_type" for col in model.cell_info_metadata.ontological_columns):
+            raise exceptions.InvalidInputError(
+                f"Model '{model.model_name}' does not have a configured ontology resource for 'cell_type'."
+            )
+        cell_type_column = next(
```
sjfleming: does this pull the first ontological_column with a column_name of "cell_type"?  can there be more than one such column?  is it always right to pull the first one?

sjfleming (approved):
I think it looks great!  One tiny question but I'm sure it's fine, I didn't see anything that looks problematic

cellarium/cas_backend/apps/compute/services/cell_operations_service.py
fedorgrab: Yeah, this is hardcoded, kinda. The thing is that client doesn't support the selection of a column. But since you mentioned that, it might be a good idea to extend the API and make "cell type" a default value. I think it's worth it!

fedorgrab: Now the API is fully ready to be able to serve multi columb processing. The client isn't updated, once it is, it will be up to user to decide.

sjfleming (approved):
Looks good, no additional suggestions from me!

Issue #212, opened 2026-04-09
Model Inference Service batch keys


fedorgrab:
Model Inference service is relying on model checkpoint to get a mapping for batch keys for input anndata. When a new model is trained on a data extract with different var or obs, client becomes incompatible. We need to implement a stable CellariumAnnDataModule that has hardcoded batch key labels compatible with the client.

PR #213, opened 2026-04-13
fix: #212 fix model inference batch keys

fedorgrab:
3 things: 
1. This change allows CAS backend to use Models without dependence on the dataloader that was used for training the model. 
2. memory budget attribute for vs index can be changed via admin
3. deployment config is changed to be 1 instance per 1 request. Now all compute allocates to one request processing.


sjfleming (approved):
Looks great

PR #219, opened 2026-07-01
Bump cellarium-ml version and update dependencies

### LINKED ISSUE #218: Update cellarium-ml in CAS backend
It seems like newly trained models don't work in CAS anymore, so we need to update the cellarium-ml package.

fedorgrab:
This PR has 2 main changes:
1. Update of cellarium ML to version 0.0.10, which required updating some other dependenceies (torch + lightning).
2. Updating Model service so that it can be compatible with scVI models.

Most of the code change is made by poetry lock.

Closes #218


sjfleming (approved):
great!

PR #221, opened 2026-07-01
feat: #220 add index creation logic

### LINKED ISSUE #220: Transfer index creation logic from cellarium-cas-workflows
Currently, vs index creation logic exist in cellarium-cas-workflows as a single monolithic function. It should be added to this repo and if needed there, used as imported module.

fedorgrab:
This pr introduces the script that creates vsindex. It was in cellarium-cas-workflows previously, but I wanted to move it here, so it can live outside of the pipeline and be just imported there.

Closes #220


cellarium/cas_backend/scripts/create_vsindex.py
```diff
@@ -0,0 +1,432 @@
...
+
+    estimated_total = int((total_training_rows / training_batches_used) * total_batches)
+    partitions = _choose_partitions(estimated_total, max_partitions)
+
+    dist_metric = _resolve_distance_metric(distance_metric)
+
+    logging.info(f"Creating IVF_FLAT index: {partitions} partitions, ~{estimated_total} total vectors")
+
+    index = vs.ingest(
+        index_type="IVF_FLAT",
```
sjfleming, line 214: out of curiosity, did you try `VAMANA` in the past?
https://cloud.tiledb.com/academy/structure/ai-ml/vector-search/foundation/data-model/#vamana

cellarium/cas_backend/scripts/create_vsindex.py
```diff
@@ -0,0 +1,432 @@
...
+    t0 = time.time()
+    logging.info("Phase 1: Loading training data …")
+
+    dfs = []
+    total_training_rows = 0
+    training_batches_used = 0
+
+    for i in range(total_batches):
+        df = _read_csv_gz(f"{embeddings_prefix}/batch_{i}.csv.gz")
+        dfs.append(df)
```
sjfleming, line 147: Don't know if you need a speedup here: you may not... but if you do, multiprocessing / multithreading (I forget which) can make this kind of multi-csv-read pretty fast

Might not be worth it at all if this is not a bottleneck

sjfleming: I've run into situations in the past where sequential reading of these was like 40 mins and parallel reading was like 5 mins

sjfleming: oooooh nevermind, you're breaking this loop after a certain size

cellarium/cas_backend/scripts/create_vsindex.py
```diff
@@ -0,0 +1,432 @@
...
+        max_partitions,
+    )
+    del ids, embeddings
+    gc.collect()
+
+    remaining_batches = list(range(training_batches_used, total_batches))
+    _stream_updates(index, embeddings_prefix, embedding_dim, remaining_batches, normalize, update_chunk_size)
+
+    if remaining_batches:
+        index = _consolidate(index, index_path, normalize)
```
sjfleming, line 380: what's this part all about?  if some batches have not been ingested into the index?  why would that happen?

sjfleming (commented):
I think it looks good.  I have just a couple comments and questions.  I had initially recommended some kind of parallel .csv.gz file reading, but I don't know if it actually makes sense in this case, since you're only reading a certain smaller number of cells upfront to create the index, and then you're (necessarily sequentially) ingesting more cells into the index after that.

cellarium/cas_backend/scripts/create_vsindex.py
fedorgrab: Never, I have no idea what this thing is. Maybe I should learn more about it!

cellarium/cas_backend/scripts/create_vsindex.py
fedorgrab: So, remaining_batches is empty, when training size is equal to the the whole data that is supposed to be injected. This case is weird, and I had this only during quick tests, when I tried to create like little indexes, without changing training batch size, then consolidate tries to consolidate, but there's nothing to consolidate, and If I recall this right, it was raising an error.

cellarium/cas_backend/scripts/create_vsindex.py
sjfleming: not necessary now, you can make a new issue!  just wondering if it might have some advantages / disadvantages compared to IVF_FLAT.  but IVF_FLAT is working fine currently.

cellarium/cas_backend/scripts/create_vsindex.py
sjfleming: ah okay, makes sense!

sjfleming (approved):
Looks good

PR #226, opened 2026-07-16
Index thinning

### LINKED ISSUE #225: Index thinning capability
# VS Index Create: add a thinning (allow-list) feature

## Summary
VS (Vector Search) Index Create should support a thinning feature. The user
can provide a path to a CSV file containing the allowed IDs — a subsample of
the original extract. During index creation, only cells whose IDs appear in
this allow-list are used; all others are filtered out.

## Allow-list contract
- The allow-list is supplied as a path to a CSV file with a single `soma_joinid`
  column (header included).
- The script does **not** verify that the CSV is genuinely a subsample of the
  provided embeddings. Providing a valid sample is the user's responsibility.
- The allow-list is used **only** for filtering — nothing else.

## Goal
The end goal is unchanged: create a VS index in which some of the data is used
for training and the remainder is used for updating. The allow-list only
changes *which* cells are eligible; it does not change the train/update
mechanism itself.

## Behavior
The update set is generated by iterating over all input batches in order and
adding vectors to the training set one by one, skipping any whose ID is not in
the allow-list. Once the training set is full, the remaining allowed cells form
the update set.

## Data-loss requirement
Filtering is intentional: cells absent from the allow-list are dropped by
design. Beyond that filtering, **no allowed cell may be lost.**

The real risk is at the batch that crosses the train/update boundary. Because
the training set is filled batch by batch, that boundary batch is only
partially consumed by training — its remaining allowed cells must be carried
into the update set, never discarded because the batch was "already used."

Stated as a testable invariant, over allowed cells only:

    train_count + update_count == (input cells present in the allow-list)

No allowed cell may end up in neither set.

## Recommendation
The per-batch allow-list membership check should be optimized. A `set` is the
recommended data structure (O(1) lookups, built once before the batch loop).
Apply it vectorized per batch (e.g. `batch_ids.isin(allowed_set)`) rather than
looping row by row, so "use a set" doesn't leave room for a slow implementation.

fedorgrab:
Add allow-list (thinning) support to create_vsindex                                                                                                          
                                                                                                                                                                             
 Lets you build a TileDB IVF_FLAT index over a subsample of an existing embedding set without regenerating embeddings. Previously create_vsindex indexed every cell in every 
 batch file; now an optional --allow-list-csv / allow_list_csv= (local path or gs://, single soma_joinid column with header) restricts indexing to the listed cells and      
 drops the rest.                                                                                                                                                             
                                                                                                                                                                             
 Behavior                                                                                                                                                                    
 - The filter is applied per batch in both phases — training (Phase 1) and streaming updates (Phase 2) — so the train/update split is preserved and every allow-listed cell  
 is indexed exactly once. Training fills to --training-sample-size allowed cells; the boundary batch is appended whole and may slightly overshoot, matching the existing     
 unfiltered path.                                                                                                                                                            
 - Phase 3 consolidation is now gated on vectors actually written rather than "remaining batches existed" — with thinning, remaining batches can filter down to zero rows.   
 Empty chunks short-circuit instead of issuing zero-row update_batch calls.                                                                                                  
 - An allow-list matching no input cell now fails fast with a ValueError instead of erroring deep inside index creation.                                                     
 - The allow-list is held as a pd.Index, not a Python set: isin() then reuses one pre-built C-level hash table instead of rebuilding a temporary object array on every       
 per-batch call — material at millions of IDs across thousands of batches.                                                                                                   
 - The script does not verify the allow-list is a genuine subsample of the embeddings; that stays the caller's responsibility.                                               
                                                                                                                                                                             
 Also in here                                                                                                                                                                
 - Extracted _pack_and_update() to de-duplicate the object-array packing + update_batch write.                                                                               
 - Tests for CLI kwarg forwarding, _load_allowed_ids parsing, boundary overshoot, the empty-allow-list error, and a stubbed end-to-end check that the train and update ID    
 sets are disjoint and together cover exactly the allow-list.                                                                                                                
 - tests/unit/scripts/conftest.py: test-only tiledb / tiledb.vector_search stub so the module imports where the native tiledb-vector-search wheel is unavailable (e.g.       
 linux/aarch64).                                                                                                                                                             
 - .gitignore entries for stray python / python3.1x artifacts; CHANGELOG entry under 1.8.5.     

Closes #225

### DIFF
--- a/cellarium/cas_backend/scripts/create_vsindex.py
+++ b/cellarium/cas_backend/scripts/create_vsindex.py
@@ -97,18 +97,24 @@ def _read_csv_gz(path: str) -> pd.DataFrame:
 
 
 @_transport_retry
-def _load_allowed_ids(path: str) -> set[int]:
+def _load_allowed_ids(path: str) -> pd.Index:
     """
-    Read the allow-list CSV (single 'soma_joinid' column, header included) into a set.
+    Read the allow-list CSV (single 'soma_joinid' column, header included) into a pd.Index.
+
+    Returns a pd.Index (int64) whose internal hash table is built once here and reused by
+    every per-batch isin() call. Passing a plain Python set to isin() forces pandas to
+    re-iterate the entire set and rebuild a temporary object array on EVERY call — for
+    millions of IDs across thousands of batches that overhead compounds to hours. A pd.Index
+    avoids that entirely: isin() routes through values._engine.ismember() which reuses the
+    pre-built C-level hash table with zero per-call setup overhead.
 
     Unlike _read_csv_gz (header-less embedding batches), this CSV has a header row so
-    pandas infers it. smart_open transparently decompresses a .gz path. The set is built
-    once here and reused for every per-batch membership check.
+    pandas infers it. smart_open transparently decompresses a .gz path.
     """
     with open(path, "rb") as f:
         buf = io.BytesIO(f.read())
     df = pd.read_csv(buf)
-    return set(df["soma_joinid"].to_numpy(dtype=np.int64).tolist())
+    return pd.Index(df["soma_joinid"].to_numpy(dtype=np.int64))
 
 
 def _normalize_in_place(arr: np.ndarray) -> None:
@@ -147,7 +153,7 @@ def _load_training_data(
     embedding_dim: int,
     training_sample_size: int,
     normalize: bool,
-    allowed_ids: set[int] | None = None,
+    allowed_ids: pd.Index | None = None,
 ) -> tuple[np.ndarray, np.ndarray, int, int, tuple[np.ndarray, np.ndarray] | None]:
     """
     Phase 1 (loading): Read batch files until training_sample_size vectors are collected.
@@ -295,7 +301,7 @@ def _process_chunk(
     embeddings_prefix: str,
     embedding_dim: int,
     normalize: bool,
-    allowed_ids: set[int] | None = None,
+    allowed_ids: pd.Index | None = None,
 ) -> int:
     """
     Read a chunk of batches and write them to the index via update_batch.
@@ -356,7 +362,7 @@ def _stream_updates(
     remaining_batches: list[int],
     normalize: bool,
     update_chunk_size: int,
-    allowed_ids: set[int] | None = None,
+    allowed_ids: pd.Index | None = None,
     carryover: tuple[np.ndarray, np.ndarray] | None = None,
 ) -> int:
     """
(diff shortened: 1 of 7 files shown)


cellarium/cas_backend/scripts/create_vsindex.py
```diff
@@ -129,40 +153,76 @@ def _load_training_data(
...
+                break
+        else:
+            # Thinning: filter by allow-list, then fill to EXACTLY training_sample_size,
+            # carrying the boundary batch's leftover allowed cells into the update set.
+            df = df[df.iloc[:, 0].isin(allowed_ids)]
+            if total_training_rows + len(df) >= training_sample_size:
+                need = training_sample_size - total_training_rows
+                train_dfs.append(df.iloc[:need])
+                carry_df = df.iloc[need:]
+                total_training_rows += need
```
sjfleming: is `need` right, or is it `len(df) - need` ?

sjfleming: oh wait, it looks like it is `need`

cellarium/cas_backend/scripts/create_vsindex.py
```diff
@@ -259,42 +333,64 @@ def _process_chunk(
...
-
-    del chunk_ids, chunk_emb, emb_obj
+    del chunk_ids, chunk_emb
     gc.collect()
 
     return chunk_rows
 
 
+@_tiledb_retry
+def _write_carryover(index: vs.IVFFlatIndex, ids: np.ndarray, emb: np.ndarray, normalize: bool) -> int:
```
sjfleming: The "carryover" concept introduced in this PR is a little complex.  I gather it's the "extra" cells opened in a file but not used to fill the initial index.  And we want to use them later as updates.

I wonder: is there a way to eliminate the need for the carryover concept entirely?  Something like:
- just allow the size of the initially-created index to be a bit flexible, so no carryover is needed

sjfleming: or:
- have a first phase where you load embedding files, keep only allowed cells in memory, and then write them out to a new file every time you get enough cells. in the new files, n_cells is controlled and you know all cells are allowed. then do what you used to do

sjfleming (commented):
Some comments about the "carryover" concept and whether it can be eliminated.  If it's necessary, that's okay.  Just looking for a potential simplification

cellarium/cas_backend/scripts/create_vsindex.py
fedorgrab: I think I like this second idea, it will make all the subsequent phases to be identical no matter if we need thinning or not. Good one, thanks!

sjfleming (approved):
Looks good to me!

PR #231, opened 2026-08-04
fix: #230 add eval mode to model inference

### LINKED ISSUE #230: Turning on eval mode in Model Inference Service
## Summary                                                                                                                                                                              
                                                                                                                                                                                           
   `ModelInferenceService` never puts the model in eval mode. `_load_module_from_checkpoint` uses                                                                                          
   `CellariumModule.load_from_checkpoint(...)` and `_get_output_from_model` calls                                                                                                          
   `cellarium_module(batch)` directly, so the module keeps `nn.Module`'s default `training=True`.                                                                                          
   Every embedding the service produces — CAS server queries and the local runner — is computed                                                                                            
   with **dropout active** and, where configured, **BatchNorm using chunk statistics instead of                                                                                            
   running stats**.                                                                                                                                                                        
                                                                                                                                                                                           
   ## The index is not affected                                                                                                                                                            
                                                                                                                                                                                           
   Index vectors come from the training pipeline's transform step, which runs under                                                                                                        
   `Trainer.predict` → `on_predict_model_eval` → `.eval()`. So `transform/batch_*.csv.gz` and the                                                                                          
   TileDB index are clean. **Only the query side is wrong; no re-indexing needed.**                                                                                                        
                                                                                                                                                                                           
   ## Evidence                                                                                                                                                                             
                                                                                                                                                                                           
   - lightning 2.5.2 `pytorch/core/saving.py::_load_from_checkpoint` ends at `return model.to(device)` (`:99`) — no `.eval()`; `CellariumModule.configure_model` doesn't call it either.   
   - cellarium-ml `layers/vae.py`: `DressedLayer` builds `nn.Dropout(p=dropout_rate)` (`:77`) and `nn.BatchNorm1d` (`:74`) into the encoder.                                               
   - cellarium-ml `models/scvi.py`: `dropout_rate` defaults to `0.1` (`:572`), injected into every encoder hidden layer (`:764-766`). `predict` returns `qz.mean` (`:1122`), so it would   
 otherwise be deterministic.                                                                                                                                                               
   - `_load_module_from_checkpoint` is `@cache`d ⇒ one shared instance per process ⇒ BatchNorm `running_mean`/`running_var` are mutated by inference and **leak across requests**.         
   - `CellariumPipeline.predict` has no `no_grad`/`inference_mode` ⇒ serving also builds an autograd graph.                                                                                
                                                                                                                                                                                           
   ## Impact                                                                                                                                                                               
                                                                                                                                                                                           
   - **PCA: none.** No mode-dependent module; train vs eval is a no-op.                                                                                                                    
   - **scVI:** identical requests return different embeddings; results depend on chunk size and on                                                                                         
     prior requests in the same process; query vectors are systematically offset from the                                                                                                  
     eval-mode index, degrading kNN retrieval. Magnitude not yet quantified.                                                                                                               
                                                                                                                                                                                           
   ## Reproduce                                                                                                                                                                            
                                                                                                                                                                                           
   1. `embed_adata` the same AnnData twice with an scVI model → embeddings differ (dropout).                                                                                               
      Identical for PCA.                                                                                                                                                                   
   2. Same input, chunk size 512 vs 8192 → extra per-cell shift for `use_batch_norm: "both"`                                                                                               
      models. Dropout noise is chunk-size-invariant, so this isolates the BatchNorm component.                                                                                             
   3. A third call may differ from both → running-stat mutation via the cached module.                                                                                                     
                                                                                                                                                                                           
   ## Fix                                                                                                                                                                                  
                                                                                                                                                                                           
   - `.eval()` on the module in `_load_module_from_checkpoint` before it is cached/returned.                                                                                               
   - Wrap `_get_output_from_model` in `torch.inference_mode()` (independent win; grad mode does                                                                                            
     not affect dropout/BatchNorm).                                                                                                                                                        
   - Optional regression guard: assert `not module.training` in `embed_adata`.                                                                                                             
                                                                                                                                                                                           
   ## Scope                                                                                                                                                                                
                                                                                                                                                                                           
   Index artifacts and PCA annotations unchanged. scVI annotations will change toward                                                                                                      
   correctness — re-run one scVI benchmark to record the delta. `cellarium-cas-workflows`' local                                                                                           
   runner inherits this path and will change with it, as intended.

fedorgrab:
Turns on inference mode and uses cellarium-ml's eval() scope. 

closes #230

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ The format is based on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`
 and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`_.
 
 
-1.8.5 - 2026-07-01
+1.8.5 - UNRELEASED
 ------------------
 
 Added
@@ -15,6 +15,10 @@ Added
 - Added ``create_vsindex`` script and CLI for building vector search indexes, moved from the external pipeline repository (#220, #221)
 - Added unit tests for the ``create_vsindex`` script
 - Added an allow-list (thinning) option to ``create_vsindex`` that filters cells by ``soma_joinid``, indexing only allowed cells while preserving the train/update split (#225)
+
+Fixed
+~~~~~
+- Fixed model inference service running query embeddings with dropout active and BatchNorm using per-chunk statistics: ``_load_module_from_checkpoint`` now calls ``.eval()`` before caching the module, and the forward pass is wrapped in ``torch.inference_mode()`` (#230)
 
 Changed
 ~~~~~~~
--- a/cellarium/cas_backend/apps/model_inference/services.py
+++ b/cellarium/cas_backend/apps/model_inference/services.py
@@ -5,6 +5,7 @@ import typing as t
 import anndata
 import numpy as np
 from smart_open import open
+import torch
 
 from cellarium.cas_backend.apps.model_inference import exceptions
 from cellarium.cas_backend.core.config import settings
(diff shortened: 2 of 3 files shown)


sjfleming (approved):
great!

